PR #9, opened 2012-09-11
Update of pom.xml and TeamFoundationServerScm.java to support "NO_CI" flag

dbascans:
- Add the ability of avoid the scm trigger when the check-in comment has the flag "NO_CI" in the comment

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,16 +1,15 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
   <modelVersion>4.0.0</modelVersion>
   <parent>
-    <groupId>org.jvnet.hudson.plugins</groupId>
+    <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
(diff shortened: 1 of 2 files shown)


olivierdagenais:

pom.xml
```diff
@@ -46,27 +61,7 @@
         </configuration>
       </plugin>
 
-      <plugin>
```
mosabua: Imho this should not be removed. If anything it should be updated to java 1.6 or 1.7 if thats what you want to support. But NOT remove. Also it should be in a separate pull request since this is not related to the actual change.

pom.xml
```diff
@@ -1,16 +1,15 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
   <modelVersion>4.0.0</modelVersion>
   <parent>
-    <groupId>org.jvnet.hudson.plugins</groupId>
```
mosabua: Are these changes required for this PR? If not they should be removed.

mosabua:
This would have to be rebased and changed quite a bit. At this stage I am not sure about the value of this feature in any case and therefore I am closing this PR. It seems weird to control CI server builds via commit messages. Feel free to submit a new PR based on the latest codebase that re-implements the feature.

PR #11, opened 2013-05-04
JENKINS-3002 TFS Support to get labels

kopcheski:
Support to get labels working now.
It's based on a build parameter called CHECKOUT_INFO which tells the plugin which strategy to be used to check files out. Without the parameter or when is empty, no special action will be done and everything may work as usual.

If there's anything I can help you with, please let me know.


olivierdagenais:
Pull request #24 was submitted and merged as a re-work of the same feature.

PR #26, opened 2014-09-19
[FIXED JENKINS-16207] TFS 2012+ Links Fixed

DD202:
Fixed the merge issues with the last request. Option exists under the Adanced section to  use the REST style urls created with TFS 2012+


olivierdagenais:

mosabua:
I would love to get this one into a release soon! +1

mosabua:
@olivierdagenais I was just checking the repo browser URL persistence on a real server and found that we need this merged for links to changesets and so on to actually work. Are you mostly okay with how this PR is implemented? If yes, I will clean it up into a separate branch and then test and merge it.

mosabua:
Also might default the new parameter to true so that it will work out of the box for new TFS server installations..

mosabua:
Given that we do not even support older versions out of the box I think this might be overkill. Instead of a config flag and so on we might want to just change the URL to use the new URLs. Can you confirm that these style URLs are used in all current versions? I will prepare a PR for this approach next.

mosabua:
Replaced by https://github.com/jenkinsci/tfs-plugin/pull/71

PR #27, opened 2014-09-23
TFS Exclusion request JENKINS-24593

sandyzden:
Fix regarding the TFS Exclusion user from SCM Trigger (just like in SVN Plugin ) : JENKINS-24593


olivierdagenais:

doronshai:
is this feature is about excluding areas that the polling is not relevant for? this is really needed.... for the git plugin there are much more options...

olivierdagenais:
[Version 4.1.0](https://github.com/jenkinsci/tfs-plugin/releases/tag/tfs-4.1.0) introduced "workspace cloaking", which allows one to exclude paths from polling and check-out.  This pull request appears to be for excluding changesets from specific users.

doronshai:
You are right, ?I tried to use the cloaking but if i have 100 folders on 1st level and each one of them have 100 folders inside, then i need to explicitly list all of these.... it would be nice to have a way to exclude recursivly a folder or to explicitly include a folder.

sandyzden:
this change is more over ignoring the SCM Trigger when a certain set of user like "build" account have committed the changes. During the build process there may be scenario to update the version files with jenkins build number and commit back to TFS repo which in turn shouldn't trigger one more jenkins run, resulting in build loop.

PR #34, opened 2015-03-16
Partial Success == Failure

mosabua:
changed to cause a partial success of a tfs operation to be considered a failure rather than a success.

See the inline javadoc for reasoning. 

If @olivierdagenais or anybody else considers this a problem, I can change it to use a global configuration (or job specific configuration) defaulting to the old behaviour.

I checked out the TFS documentation as well as the commit history of this plugin but I could not find any indication why a partial success is considered a success. If anybody has more insight I would love to know.

### DIFF
--- a/src/main/java/hudson/plugins/tfs/TfTool.java
+++ b/src/main/java/hudson/plugins/tfs/TfTool.java
@@ -103,8 +103,16 @@ public class TfTool {
         
         int result = proc.join();
         LOGGER.fine(String.format("The TFS command '%s' returned with an error code of %d", toolArguments[1], result));
-        if ((result == SUCCESS_EXIT_CODE) || (result == PARTIAL_SUCCESS_EXIT_CODE)) {
+        if (result == SUCCESS_EXIT_CODE) {
             return new InputStreamReader(new ByteArrayInputStream(consoleStream.toByteArray()));
+        } else if (result == PARTIAL_SUCCESS_EXIT_CODE) {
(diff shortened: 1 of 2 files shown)


cgnuechtel:
Maybe TFS "partial success" better maps to Jenkins "unstable" state!?

mosabua:
Well .. I am not sure. I think a partial success is completely a failure. At least we had cases where a checkout of a project was a "partial success" yet it e.g. checked out the project but did not traverse all the way downs in to some java package directories (path too long). Since it was considered a success up until this PR the build proceeded with only parts of the codebase. Imho it should have never even proceeded to the next build step. I am not aware of a situation where a "partial" success makes sense at all.

cgnuechtel:
@mosabua: Yes, you are absolutely right. In the context of source control checkout, partial success does not make any sense. So, please forget my comment.

(I had TFS test execution in mind, which does not matter here. If tests fail on TFS, the TFS build result is "Partial success".)

mosabua:
With the changes to using the embedded TFS SDK this PR no longer applies. We should however test and confirm that a checkout with excessive path length does result in a failure rather than a "partial success" or whatever and then proceeds with the build.

mosabua:
I think this issue https://issues.jenkins-ci.org/browse/JENKINS-30241 shows that long paths now are no longer shadowed and actually result in a build failure as expected. The error message is okay as well imho. @olivierdagenais I think you can close this PR but will leave it up to you.

PR #35, opened 2015-03-17
overwrite flag

mosabua:
added advanced job config option to pass in the overwrite flag to the tfsvc get command:
- is an advanced job level config option
- defaults to false

I would like to add others such as all and force but I am not sure about the proliferation of parameters. Maybe instead of doing the current overwrite flag we should introduce some sort of "parameters" input where people can just provide things like "/overwrit /all /force" or whatever.

It might also be good to have a global config default.


mosabua:
wdyt @olivierdagenais

apieraf:
I would like to have this feature in the next version of the plugin, thanks

mosabua:
This would have to be heavily refactored to be used. Not even sure if the current SDK usage even needs the flag. For now will leave the PR open as a reminder but I have not objections to close it if so desired.

mosabua:
I am closing this since it is essentially useless compared to the current master.

PR #24, opened 2014-06-08
JENKINS-3002 TFS Support to get labels (2nd attempt)

kopcheski:
Implemented the ability to checkout by label. 
https://issues.jenkins-ci.org/browse/JENKINS-3002

This pull request overtakes the last one of mine, with much less code changes.

PR #36, opened 2015-04-01
Fix checkout by version spec

olivierdagenais:
Two defects were discovered in the implementation of pull request #24:

1. The value of `VERSION_SPEC` was incorrectly applied.
2. The reported SCM changes between builds by version spec was incorrectly querying TFVC.

Outstanding weirdness/defect:
-------------------------------

I created a job with polling, then added the `VERSION_SPEC` parameter with a default value of `LFoo` (a label that pointed to changeset 4).  Running a build indeed got the source from changeset 4, but then subsequent polling would detect that a changeset newer than 4 existed and queue a build.  The build would again obtain changeset 4 and the loop repeated itself until I turned off polling or removed the default value.

Part of the problem was due to configuration error (it probably doesn't make sense to supply a default value to a `VERSION_SPEC` parameter - I only did it because I'm testing the feature), but a configuration error should NOT cause a build loop. The loop itself is difficult to cause for most users, since the "checkout by label" feature isn't [yet] documented nor easily discoverable, so it might not be a very big deal.

### DIFF
--- a/src/main/java/hudson/plugins/tfs/TeamFoundationServerScm.java
+++ b/src/main/java/hudson/plugins/tfs/TeamFoundationServerScm.java
@@ -199,21 +199,7 @@ public class TeamFoundationServerScm extends SCM {
             build.addAction(workspaceConfiguration);
             VariableResolver<String> buildVariableResolver = build.getBuildVariableResolver();
             String singleVersionSpec = buildVariableResolver.resolve(VERSION_SPEC);
-            int buildChangeset;
-            try {
-                setWorkspaceChangesetVersion(null);
-                String projectPath = workspaceConfiguration.getProjectPath();
-                Project project = server.getProject(projectPath);
-                // TODO: even better would be to call this first, then use the changeset when calling checkout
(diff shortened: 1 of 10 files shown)


olivierdagenais:
@TomaszDom and @kopcheski, do you want to weigh in on the subject?  @mosabua has already gone through this pull request's commits once, so I would appreciate hearing from users who are using (or want to use) the "checkout by label" feature.  *Caveat emptor*: if you want to test the plugin built by this branch, be aware that while an upgrade will work fine, a downgrade (back to the previously stable version of the TFS plugin) will result in having to re-enter the TFS user's password for all jobs.

mosabua:
+1 for merging this. I will do some further testing on this today, but initial usage seems everything is fine.

olivierdagenais:
> I will do some further testing on this today, but initial usage seems everything is fine.

@mosabua: Can you report on your soak test?

mosabua:
I got pulled away with other issues after local smoke testing. That went fine. I am hoping to do some further testing on the staging server next week. I would suggest to merge away thoug .. and then fix prior to a release if we find anything.

mosabua:
Nice!

mosabua:
Should delete the branch probably..

PR #46, opened 2015-08-13
Replace the use of `tf` (command-line client) in *Command implementations.

olivierdagenais:
This pull request removes the need to install the TEE CLC or Visual Studio Team Explorer on every agent, adding `tf` to the PATH, etc.

Testing
-------
Integration tests were run on TFS 2012, 2013 and 2015, after removing the TEE CLC from the corresponding agents.


olivierdagenais:
Odd, the CloudBees automated build did not register the success of [build 168](https://jenkins.ci.cloudbees.com/job/plugins/job/tfs-plugin/168/).

PR #61, opened 2015-11-10
Replace BuildVariableResolver with BuildEnvironment variables

aheubusch:
The resolver provided NODE_NAME, JOB_NAME and USER_NAME (in some circumstances). Since (around) jenkins core 1.305, NODE_NAME and JOB_NAME are available through the build environment variables. I am not sure what purpose USER_NAME has, and why this plug should provide it. Removed it.

I am pretty new to jenkins (and java). If someone could test this (besides me) before it gets merged that would be appreciated.


Ullio:
I used your modification for some time and it worked well. As it makes consuming build variables really easy and as I have the need to use them for the local workfolder too, I merged your changes to the *support-variables-workfolder* branch having the support for cloaked paths and shelvesets.

PR #64, opened 2015-12-07
Ignored Cloaked Paths when Polling

watsonlu:
This change makes it so that changes to cloaked paths do not trigger builds as requested by @mahsoud. This change makes pull request #60 obsolete as it includes all of the changes @ajalexander made.

The Shelveset change from #63 does not merge cleanly with this fix (its a small conflict). If you want both shelvesets + this change, I've merged the two branches here. Feel free to build it from here:

https://github.com/watsonlu/tfs-plugin/tree/MergeShelvesetAndCloakedPaths

### DIFF
--- a/src/main/java/hudson/plugins/tfs/TeamFoundationServerScm.java
+++ b/src/main/java/hudson/plugins/tfs/TeamFoundationServerScm.java
@@ -306,9 +306,10 @@ public class TeamFoundationServerScm extends SCM {
         } else {
             Server server = createServer(launcher, listener, lastRun);
             try {
-                return (server.getProject(getProjectPath(lastRun), getCloakPaths(lastRun)).getDetailedHistory(
+                return (server.getProject(getProjectPath(lastRun), getCloakPaths(lastRun)).getDetailedHistoryWithoutCloakedPaths(
                             lastRun.getTimestamp(), 
-                            Calendar.getInstance()
+                            Calendar.getInstance(),
+                            getCloakPaths(lastRun)
                         ).size() > 0);
             } finally {
                 server.close();
@@ -528,7 +529,7 @@ public class TeamFoundationServerScm extends SCM {
         final Server server = createServer(localLauncher, listener, build);
         final Project tfsProject = server.getProject(projectPath, cloakPaths);
         try {
-            final ChangeSet latest = tfsProject.getLatestChangeset();
+            final ChangeSet latest = tfsProject.getLatestUncloakedChangeset(cloakPaths);
             final TFSRevisionState tfsRemote =
                     (latest != null)
                     ? new TFSRevisionState(latest.getVersion(), projectPath)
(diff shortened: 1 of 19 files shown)


olivierdagenais:
Thank you for the pull request!  I am reviewing the changes and I appreciate the small, focused commits.

In the meantime, can you update the pull request description to include the manual testing performed to verify the feature?  In other words, what would the documentation on how to use this look like?

olivierdagenais:
I'm currently rebasing the changes and amending the commits to apply minor fixes, when, during testing, I realized that the `;` separator for cloaked paths seems more complicated than using only newlines, because I could see someone typing the following, which will pass validation, but won't do what the user expected:

```
$/BuildTemplates;
$/OldStuff
/OtherOldStuff
```

In other words, the second "cloaked path" contains a newline, instead of there being three paths.  Since paths in TFS should never contain newlines, how about we just make the "Cloaked paths" field accept its entries one per line instead?

ajalexander:
@olivierdagenais Regarding the separator character, using just a newline seems fine. The only reason I'd initially started with the `;` character was that it was more in-keeping with an [old patch](https://issues.jenkins-ci.org/browse/JENKINS-4542) that provided similar functionality.

ajalexander:
@olivierdagenais Regarding a description of how the feature works:

This patch adds the ability to cloak paths to prevent them from being fetched during a checkout. Further, it prevents check-ins to cloaked paths from triggering builds.

As an example, consider a project with the following folders:

> $/MyProject
> $/MyProject/A
> $/MyProject/A/1
> $/MyProject/A/2
> $/MyProject/B
> $/MyProject/C

The patch allows a job configuration to only checkout a subset of the folders. In this case, the main path would be setup as `$/MyProject`. If the following paths were entered in the cloaking definition:

> $/MyProject/A/2
> $/MyProject/B

then the resulting workspace on the Jenkins server would only have the following folders:

> $/MyProject
> $/MyProject/A
> $/MyProject/A/1
> $/MyProject/C

To test/verify the behavior, you can examine the actual folder structure on disk in the workspace on the Jenkins server. Checking-in a file on an uncloaked path should trigger a build. Checking-in a file on a cloaked path should not trigger a build.

olivierdagenais:
@ajalexander Thank you, I'll work that into the documentation and I'll try to craft an integration test that matches your example.

I have created the [tfs-plugin/support_workspace_cloaking](https://github.com/jenkinsci/tfs-plugin/commits/support_workspace_cloaking) branch where I have made minor changes to the work in this pull request, plus a few fixes based on my findings while reviewing.

olivierdagenais:
Just an update: while testing these changes, some of my integration tests were failing.  It looks like it might be due to a combination of a defect in the TFS SDK for Java and a change in VSTS, since I was able to reproduce on the master branch.  I want to chase that down and resolve it to avoid breaking anything further.

olivierdagenais:
Obsoleted by #68.

PR #65, opened 2015-12-25
Implemented ShowWorkspaces flag

AndreyAlifanov:
In our big project we have huge amount of workspaces.
And all of them is shown in Jenkins log - it's absolutely unusefull info for me.
But I think it may be usefull for someone else.
So, I implemented flag "ShowWorkspaces" to control log size.

May be it's implemented in not a very good manner - I'm not Java programmer.

### DIFF
--- a/src/main/java/hudson/plugins/tfs/TeamFoundationServerScm.java
+++ b/src/main/java/hudson/plugins/tfs/TeamFoundationServerScm.java
@@ -83,6 +83,7 @@ public class TeamFoundationServerScm extends SCM {
     private /* almost final */ Secret password;
     private final String userName;
     private final boolean useUpdate;
+    private final boolean showWorkspaces;
     
     private TeamFoundationServerRepositoryBrowser repositoryBrowser;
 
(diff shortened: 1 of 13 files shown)


AndreyAlifanov:
As I realized, my pull request resolves issue: JENKINS-30330.

mosabua:
Nice! I like this and we have the same problem with the performance impact of creating that list. @olivierdagenais

mosabua:
My only input would be to default to false instead of true.

AndreyAlifanov:
I'm agree about default = 'false'.
But I didn't change default because I don't know background of this plugin.

AndreyAlifanov:
What should I do to merge this pull request into master?

mosabua:
@olivierdagenais is going to take care of further review, merge and release. Stay tuned @AndreyAlifanov

PabloThiele:
Hello folks, is possible to at least have the pull of the fixes there.

olivierdagenais:
@AndreyAlifanov, @PabloThiele and anybody else experiencing this issue: would you mind taking a look at pull request #95 and testing the linked tfs-4.1.1-SNAPSHOT.hpi file on your system(s) to see if you can still reproduce the defect (all workspaces are listed)?

With commit 3964078848c00575efc72945d71744bad9297e55, only the current computer's workspaces should be listed.  Once I get confirmation that the list is indeed shorter, I'll also hide it by default.

Thanks!
\- Oli

olivierdagenais:
Obsoleted by #95.

PR #70, opened 2016-01-30
Import and improve wiki content into README.md

olivierdagenais:
The Jenkins Confluence instance is old, slow and clunky, often saving wiki markup but not re-rendering it.  Having the content in source control will enable us to branch it, edit it offline, review it, merge it and persist changes that shouldn't be published right away, such as a change in features/screenshots that will only be valid for the next release.

I have imported the content from the wiki page (which I will replace with a link to this file once this is merged), converting Textile to Markdown as closely as possible.  The release notes for previous versions have been copy-pasted into the corresponding tags in GitHub's [Release](https://github.com/jenkinsci/tfs-plugin/releases) page.  The "Present" section can be atomically updated with the corresponding changes and should be copy-pastable into the next release's notes.


olivierdagenais:
You can see what the new README looks like by browsing it in its branch: https://github.com/jenkinsci/tfs-plugin/blob/import_readme/README.md

aheubusch:
LGTM

mosabua:
+1 .. excellent idea. In the long run we could even do a full gh-pages site but I think this is good for now. Much better than the wiki anyway. I will merge this. We can refactor more when necessary. E.g. I might add a ToC at the top somewhere.

mosabua:
Do you want me to rip out the wiki page and just add a link to the readme @olivierdagenais ?

mosabua:
Also in terms of the changelog we could just import that into a single additional file. E.g. I do that on the android-maven-plugin. Although that has a long history so there are actually three files and it is rendered via Maven site and deployed to gh-pages. See http://simpligility.github.io/android-maven-plugin/changelog.html or  https://github.com/simpligility/android-maven-plugin/blob/master/src/site/asciidoc/changelog.adoc

olivierdagenais:
> Do you want me to rip out the wiki page and just add a link to the readme @olivierdagenais ?

Yes, and please keep the `{excerpt}` and that fancy "Plugin Information" table that's generated via the Confluence macro.

olivierdagenais:
> Also in terms of the changelog we could just import that into a single additional file.

GitHub already has that information at the [/releases](https://github.com/jenkinsci/tfs-plugin/releases) page.  I even imported the content from the really old releases, so we should be good to go.

mosabua:
Hahah @olivierdagenais  .. I was just about to remove the content to find you beat me to it. Sorry for the delay.

PR #71, opened 2016-02-11
[JENKINS-16207] Repo browser links fixed

mosabua:
# Description

Team Foundation Server (TFS) drastically changed the URLs to changesets, file contents and file diffs in TFS 2012, breaking links from Jenkins.  In TFS 2013 and 2015, the links generated by the web UI are slightly different but TFS 2012-style links are compatible.

Replaces pull request #26.

## Why not use the TFS SDK for Java's feature for building hyperlinks?

The `TSWAHyperlinkBuilder` class, despite not having a method for generating a link to a file diff, was used for a prototype.

Although it worked, with an overhead of _at least_ one remote procedure call (RPC) _per hyperlink_, the performance was deemed unacceptable.

# Documentation

Commit 3619ef3742f18f89aba6f12b34bd8013286c185d was added to the `documentation_4.1.0` branch. 

Here's a screenshot of the results of updating `src/main/webapp/browsers/tswa.html`, seen in context and with sample values:

Screenshot of Team Foundation Server SCM with Repository browser configured to "Team System Web Access"

# Manual testing

1. Upload private build of TFS plugin HPI to the Jenkins master via `/jenkins/pluginManager/advanced`, _Prepare for Shutdown_ via `/jenkins/quietDown` then restart Jenkins.
2. For each of TFS 2012, TFS 2013 and TFS 2015: 
    1. Create a Jenkins job to use the "Team Foundation Server" SCM with the following specifications (see the screenshot above for the values that were used for the TFS 2012 run):
        1. Set the _Server URL_ to use the TFS server's "short" host name (i.e. using WINS resolution).
        2. Set the _Repository browser_ to be **Team System Web Access** with its _URL_ set to the FQDN of the TFS server. 
    2. Launch the job so it can record a baseline changeset in the TFVC repository.
    3. Go make one or more changesets in the TFVC repository (such as by running this project's functional tests).
    4. Launch the job again and wait for it to finish.
    5. Visit the job's **Changes** page and click a sample of the **Team System Web Access** links to visit TFS's view of the changesets.
    6. Visit the last build's **Changes** page and click a sample of the _Version_ links, some changeset item links and some diff links to visit them in TFS.  Verify the results are expected, such as being consistent with changes made in a certain changeset.

Mission accomplished!

### DIFF
--- a/src/main/java/hudson/plugins/tfs/browsers/TeamSystemWebAccessBrowser.java
+++ b/src/main/java/hudson/plugins/tfs/browsers/TeamSystemWebAccessBrowser.java
@@ -90,10 +90,6 @@ public class TeamSystemWebAccessBrowser extends TeamFoundationServerRepositoryBr
         }
         return createChangeSetItemLink(item, "compare");
     }
-    
-    private String getPreviousChangeSetVersion(ChangeSet changeset) throws NumberFormatException {
-        return Integer.toString(Integer.parseInt(changeset.getVersion()) - 1);
-    }
 
     @Extension
     public static final class DescriptorImpl extends Descriptor<RepositoryBrowser<?>> {
--- a/src/main/java/hudson/plugins/tfs/browsers/TeamSystemWebAccessBrowser.java
+++ b/src/main/java/hudson/plugins/tfs/browsers/TeamSystemWebAccessBrowser.java
@@ -70,6 +70,7 @@ public class TeamSystemWebAccessBrowser extends TeamFoundationServerRepositoryBr
         final URL changeSetUrl = getChangeSetLink(changeSet);
         final QueryString qs = new QueryString();
         qs.put("path", item.getPath());
+        qs.put("version", changeSet.getVersion());
         qs.put("_a", action);
         return new URL(changeSetUrl, "#" + qs.toString());
     }
--- a/src/main/java/hudson/plugins/tfs/browsers/TeamSystemWebAccessBrowser.java
+++ b/src/main/java/hudson/plugins/tfs/browsers/TeamSystemWebAccessBrowser.java
@@ -46,7 +46,7 @@ public class TeamSystemWebAccessBrowser extends TeamFoundationServerRepositoryBr
     private String getBaseUrlString(ChangeSet changeSet) throws MalformedURLException {
         String baseUrl;
         if (url != null) {
-            baseUrl = url;
+            baseUrl = normalizeToEndWithSlash(new URL(url)).toString();
         } else {
             String scmUrl = getServerConfiguration(changeSet);
             baseUrl = normalizeToEndWithSlash(new URL(scmUrl)).toString();
--- a/src/main/java/hudson/plugins/tfs/browsers/TeamSystemWebAccessBrowser.java
+++ b/src/main/java/hudson/plugins/tfs/browsers/TeamSystemWebAccessBrowser.java
@@ -6,6 +6,7 @@ import hudson.model.AbstractProject;
 import hudson.model.Descriptor;
 import hudson.plugins.tfs.TeamFoundationServerScm;
 import hudson.plugins.tfs.model.ChangeSet;
+import hudson.plugins.tfs.util.QueryString;
 import hudson.scm.EditType;
 import hudson.scm.RepositoryBrowser;
 import hudson.scm.SCM;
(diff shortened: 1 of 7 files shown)


src/main/java/hudson/plugins/tfs/browsers/TeamSystemWebAccessBrowser.java
```diff
@@ -53,38 +53,38 @@ private String getBaseUrlString(ChangeSet changeSet) throws MalformedURLExceptio
...
             return null;
-        }
-        try {
-            return new URL(String.format("%sdiff.aspx?opath=%s&ocs=%s&mpath=%s&mcs=%s", 
-                    getBaseUrlString(parent), 
-                    item.getPath(),
-                    getPreviousChangeSetVersion(parent), 
-                    item.getPath(),
-                    parent.getVersion()));
-        } catch (NumberFormatException nfe) {
```
mosabua: version is just a string now so a NFE never happens.

src/main/java/hudson/plugins/tfs/browsers/TeamSystemWebAccessBrowser.java
```diff
@@ -104,8 +104,14 @@ public String getDisplayName() {
         }
         
         public static String getBaseUrl(String urlExample) throws MalformedURLException {
-        	URL url = new URL(urlExample);
-        	return new URL(url.getProtocol(), url.getHost(), url.getPort(), String.format("/%s", FilenameUtils.getPath(url.getPath()))).toString();
+
+          URL url = new URL(urlExample);
+          String path = url.getPath();
+          // path needs to end with / and users sometimes forget so we add it
+          if (!path.endsWith("/")) {
```
mosabua: This is necessary to avoid user annoyance. There was a bug and there are earlier tests for this.

mosabua:
@olivierdagenais okay for me to merge this to master before the release. The current implementation does not work at all from all I can tell, but this works from my testing. I do not believe we need to be able to support the old URL pattern since we do not support those old versions of TFS anyway.

mosabua:
Btw. this does not do any magic determination of the right collection or anything like that. The repo browser URL just has to be configured correctly for the specific project.

mosabua:
Added updated help text to the PR now. A colleague of mine ran this commit through the integration test suite as well. Imho this is good to merge. Wdyt @olivierdagenais

src/main/java/hudson/plugins/tfs/browsers/TeamSystemWebAccessBrowser.java
```diff
@@ -47,44 +47,49 @@ private String getBaseUrlString(ChangeSet changeSet) throws MalformedURLExceptio
         if (url != null) {
             baseUrl = DescriptorImpl.getBaseUrl(url);
         } else {
-            baseUrl = String.format("%s/", getServerConfiguration(changeSet));
```
mosabua: This caused duplicate / usage in URLs.

src/main/webapp/browsers/tswa.html
```diff
@@ -1,12 +1,11 @@
 <div>
   <p>
-	Enter the URL to the TSWA server. Examples:<tt>http://tswasrv</tt> or <tt>http://TFSMachine:8080/</tt>.
-	The plugin will strip the last path (if any) of the URL when building URLs for change set pages and other pages. 
+	Enter the URL to your project on the TFS server, used for browser-based access.
   </p>	 
   <p>
-    If the field is empty then the plugin will try to build up a valid URL using the server URL.
```
olivierdagenais: I prefer the original wording and placement of this sentence.

src/main/webapp/browsers/tswa.html
```diff
@@ -1,12 +1,11 @@
 <div>
   <p>
-	Enter the URL to the TSWA server. Examples:<tt>http://tswasrv</tt> or <tt>http://TFSMachine:8080/</tt>.
-	The plugin will strip the last path (if any) of the URL when building URLs for change set pages and other pages. 
+	Enter the URL to your project on the TFS server, used for browser-based access.
```
olivierdagenais: We should be more specific as to why someone would need to provide a value here.  In other words, when is the default/derived value not acceptable?

src/main/java/hudson/plugins/tfs/browsers/TeamSystemWebAccessBrowser.java
```diff
@@ -47,44 +47,49 @@ private String getBaseUrlString(ChangeSet changeSet) throws MalformedURLExceptio
         if (url != null) {
             baseUrl = DescriptorImpl.getBaseUrl(url);
         } else {
-            baseUrl = String.format("%s/", getServerConfiguration(changeSet)); 
+            String scmUrl = getServerConfiguration(changeSet);
+            if (scmUrl.endsWith("/")) {
+              baseUrl = scmUrl;
+            } else {
+              baseUrl = String.format("%s/", scmUrl);
+            }
```
olivierdagenais: This "ensureTrailingSlash" logic is duplicated in `DescriptorImpl#getBaseUrl` below. Consider calling that method or extracting a helper that both sites call.

src/main/java/hudson/plugins/tfs/browsers/TeamSystemWebAccessBrowser.java
```diff
@@ -47,44 +47,49 @@ private String getBaseUrlString(ChangeSet changeSet) throws MalformedURLExceptio
...
+              baseUrl = String.format("%s/", scmUrl);
+            }
         }
         return baseUrl;
     }
 
     /*
-     * http://tswaserver:8090/cs.aspx?cs=99
+     * Gets the link to a specific change set.
+     * E.g. http://tswaserver:8090/_versionControl/changeset/99
```
olivierdagenais: Would you mind updating the example URLs (and test cases, if possible) to look like the real-world values (and avoid TSWA-style URLs)?
For example: `https://vsts.example.com/DefaultCollection/` and `http://tfs.example;com:8080/tfs/DefaultCollection/`

src/main/java/hudson/plugins/tfs/browsers/TeamSystemWebAccessBrowser.java
```diff
@@ -47,44 +47,49 @@ private String getBaseUrlString(ChangeSet changeSet) throws MalformedURLExceptio
...
     /*
-     * http://tswaserver:8090/view.aspx?path=$/Project/Folder/file.cs&cs=99
+     * Gets the link for a specific file in a change set.
      */
     public URL getFileLink(ChangeSet.Item item) throws IOException {
-        return new URL(String.format("%sview.aspx?path=%s&cs=%s", getBaseUrlString(item.getParent()), item.getPath(), item.getParent().getVersion()));
+      return new URL(String.format("%s_versionControl/changeset/%s#path=%s&_a=contents",
+                                   getBaseUrlString(item.getParent()),
+                                   item.getParent().getVersion(),
+                                   URLEncoder.encode(item.getPath(),"UTF-8")));
```
olivierdagenais: Can we do better than string concat/format for URL path and query string components?  That way, we wouldn't have to worry about calling whether we need to call `URLEncoder#encode` or not; it would be taken care of by some `UrlBuilder` of some sort.  Or build the URL path with one class and the query string with another, using something like https://github.com/Microsoft/Git-Credential-Manager-for-Mac-and-Linux/blob/master/src/main/java/com/microsoft/alm/helpers/UriHelper.java#L30

mosabua:
@olivierdagenais  Feel free to update what is in the branch to your liking. I can only look at this in about 2 weeks time..

mosabua:
Thanks @olivierdagenais for the ton of improvements on top of my hack ;-)

PR #74, opened 2016-03-22
Updated repo browser label 

mosabua:
to be up to date with product names and in sync with git plugin

See https://github.com/jenkinsci/git-plugin/pull/387

Tested this on local run and test deployment.


mosabua:
I can also start a new release doc section for 4.1.1 and add there, if desired.

And in terms of the documentation I am not sure if we need a new branch for 4.1.1 docs and still have a merge outstanding or what process you are using around that @olivierdagenais

mosabua:
The git plugin change is merged

olivierdagenais:
I would prefer documentation changes go in their own branch so the wiki doesn't look out of skew with the current latest release, but in this case it's not a big deal and there is usually just one choice of repo browser.

mosabua:
OKay... I will keep doc changes separate in the future.

PR #79, opened 2016-06-01
Cloaked paths should be case insensitive when polling for changes

watsonlu:
We noticed this issue last night. If you add a cloaked path like " $/Server/TeSt2" when the actual folder name on the server is "test2", a build will still be triggered if someone checks into that folder. The polling is case sensitive.

This is a quick fix for this case, polling is now insensitive.

No changes are needed to the get, it's already case insensitive. Polling now matches this behavior.

### DIFF
--- a/src/main/java/hudson/plugins/tfs/model/Project.java
+++ b/src/main/java/hudson/plugins/tfs/model/Project.java
@@ -234,7 +234,7 @@ public class Project {
         for (final String tfsPath : changesetPaths) {
             boolean isPathCloaked = false;
             for (final String cloakedPath : cloakedPaths) {
-                if (tfsPath.toLowerCase().startsWith(cloakedPath.toLowerCase())) {
+                if (tfsPath.regionMatches(true, 0, cloakedPath, 0, cloakedPath.length())) {
                     isPathCloaked = true;
                     break;
                 }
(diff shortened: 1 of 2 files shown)


olivierdagenais:
Thank you for the pull request!

I'm always weary of calling case-conversion methods on a pair of strings to perform a case-insensitive comparison because I know there are characters that only exist in one case (like the [German "sharp S"](https://en.wikipedia.org/wiki/%C3%9F)), and there also exist non-standard case conversions (like the [Turkish dotted and dotless I](https://en.wikipedia.org/wiki/Dotted_and_dotless_I)).  For that reason, I prefer to defer to framework-provided string comparison methods that have a case-insensitive option.  For our situation, it appears [`String#regionMatches`] (http://docs.oracle.com/javase/7/docs/api/java/lang/String.html#regionMatches%28boolean,%20int,%20java.lang.String,%20int,%20int%29) would do the trick.

To make sure we get this right and don't risk breaking it in the future, can I trouble you to also add a test to `ProjectTest`?  There are already a number of tests for `Project#isChangesetFullyCloaked`; it should be easy enough to add a new one.

Thanks!
\- Oli

watsonlu:
Thanks for the feedback Olivier,

I've swapped to using regionMatches like you suggested and added a test for it

olivierdagenais:
Thank you for making the suggested changes!

PR #81, opened 2016-07-12
Upgrade Jenkins dependencies to 1.580

olivierdagenais:
Most of the resulting modifications were test-related.

Manual testing
--------------

1. `mvn verify clean`
2. Install `target/tfs.hpi` in a Jenkins instance.
3. Click around to configure a job that uses the TFS plugin, hit Save and go back to make sure the values are still there.
4. Queue a build that pull sources from TFVC.

Initial smoke tests pass!  The one thing that no longer appears to work is running through `mvn hpi:run` and that appears to be a problem with the classpath.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>1.448</version>
+    <version>1.580</version>
   </parent>
 
   <artifactId>tfs</artifactId>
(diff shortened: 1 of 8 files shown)


mosabua:
Btw @olivierdagenais mvn hpi:run works fine for me. Try mvn clean install hpi:run in the extreme case ..

It runs at http://localhost:8080/jenkins/ after it fully starts

mosabua:
But I am getting a jelly tag exception when trying to configure a job 

Caused by: java.lang.NoClassDefFoundError: com/microsoft/tfs/core/clients/versioncontrol/specs/version/VersionSpec

so maybe there are some issues of picking up the tfs library from the web-inf folder with hpi:run 

Imho we should upgrade to a newer plugin parent anyway.. but this is a first step ;-)

olivierdagenais:
> But I am getting a jelly tag exception when trying to configure a job

That's what I meant by "`mvn hpi:run` no longer works".  I even tried to set the `pluginFirstClassLoader` configuration option for the `maven-hpi-plugin` without success.

I was able to work around the difficulty if I shuffled the help files, replaced the `field` attribute with a `name` attribute in the Jelly files, etc. and added a `help` attribute where appropriate but that did not seem sustainable and more of a hack.

Let the record show that I'll buy a $BEVERAGE to whoever can solve this classpath mystery. :)

PR #82, opened 2016-07-12
Add David to the list of developers in pom.xml

DavidStaheli:


olivierdagenais:
With great power comes great responsibility.

PR #62, opened 2015-11-10
Change minimum run time requirement to Java6

aheubusch:
Title speaks for itself i guess. jenkins-core started using java6 mid 2013 and java7 mid 2015.
http://jenkins-ci.org/content/good-bye-java6

PR #83, opened 2016-07-13
changed language level check to use 1.6

mosabua:
A side effect of https://github.com/jenkinsci/tfs-plugin/pull/81 is that the language level for Jenkins plugin compilation (and Jenkins) changes to Java 6. 

This PR fixes the build check to validate java 6 instead of java 5. This essentially allows Java 6 usage without breaking the build (not that there were any language changes but anyway...) .

Ran full build and tested deployment and usage of plugin manually. No IT runs necessary since build output is not affected by this change.

This PR and the above merged PR replace https://github.com/jenkinsci/tfs-plugin/pull/62, which I will close.

Unless there are any objections I will merge this tomorrow. fyi @olivierdagenais

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -127,6 +127,7 @@
       <plugin>
         <groupId>org.codehaus.mojo</groupId>
         <artifactId>animal-sniffer-maven-plugin</artifactId>
+        <!-- version inherited from parent pom -->
         <configuration>
           <signature>
             <groupId>org.codehaus.mojo.signature</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -125,27 +125,25 @@
         </configuration>
       </plugin>
       <plugin>
-	<groupId>org.codehaus.mojo</groupId>
-	<artifactId>animal-sniffer-maven-plugin</artifactId>
-	<version>1.7</version>
-	<configuration>
-	  <signature>
-	    <groupId>org.codehaus.mojo.signature</groupId>
-	    <artifactId>java15</artifactId>
-	    <version>1.0</version>
-	  </signature>
-	</configuration>
-	<executions>
-	  <execution>
-	    <id>java.1.5.check</id>
-	    <phase>verify</phase>
-	    <goals>
-	      <goal>check</goal>
-	    </goals>
-	  </execution>
-	</executions>
+        <groupId>org.codehaus.mojo</groupId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -125,27 +125,25 @@
         </configuration>
       </plugin>
       <plugin>
-	<groupId>org.codehaus.mojo</groupId>
```
olivierdagenais: Let the record show there used to be a mix of TAB and spaces, which is why the whole block appears changed.

pom.xml
```diff
@@ -125,27 +125,25 @@
...
-	<artifactId>animal-sniffer-maven-plugin</artifactId>
-	<version>1.7</version>
```
olivierdagenais: I will insist we keep the plugin version, to make builds more predictable.

olivierdagenais:
An easy enough verification of this change would be to try using `@Override` on an interface implementation or to try using a method introduced in 1.6, such as `File#setWriteable`.  Either should fail the build without this change and succeed with it.

olivierdagenais:
If you don't mind anchoring the `org.codehaus.mojo:animal-sniffer-maven-plugin` version before merging (it doesn't have to stay at version 1.7), that would be great (you should be able to amend the commit).

BTW, the push of your branch ran the integration tests on my VMs and everything was fine.  I'll look into surfacing this status in GitHub commits and/or pull requests.

Thank you for your contribution!

pom.xml
mosabua: The plugin version is inherited from the parent and overriding it here with an older version is bad and does NOT make it more reproducible. If anything you want to override it with a newer version .. but generally this should be inherited from the parent.,

pom.xml
mosabua: Confirmed... tabs have to go ;-)

pom.xml
olivierdagenais: > The plugin version is inherited from the parent

Oops, I missed that fact.  How about we replace the `<version>` line with something like the following?
`<!-- version is inherited -->`

mosabua: Seems redundant but if it helps .. sure.

mosabua:
I did some Java 6 checks as suggested manually and it all worked as expected. None of those are committed though..

PR #84, opened 2016-07-13
Automatically configure used webbrowser access

mosabua:
The updated core dependency to 1.580 allows us to override the new guessBrowser method. It sets a default web access browser if none is configured on a job.

Since there is only one TFVC web browser implementation we can assume it to be used and therefore activated when "(Auto)" e.g. nothing , is configured in the SCM sections "Repository browser"	drop down.

Full build is passing. 

Tested with a Jenkins 2.7.1 instance and a TFS server as well as a VS Online account. Works as expected.

### DIFF
--- a/src/main/java/hudson/plugins/tfs/TeamFoundationServerScm.java
+++ b/src/main/java/hudson/plugins/tfs/TeamFoundationServerScm.java
@@ -421,9 +421,8 @@ public class TeamFoundationServerScm extends SCM {
     
     @Override
     public @CheckForNull RepositoryBrowser<?> guessBrowser() {
-      return new TeamSystemWebAccessBrowser( serverUrl + projectPath );
+      return new TeamSystemWebAccessBrowser(serverUrl);
     }
-    
 
     // Convenience method for tests.
     public void setRepositoryBrowser(final TeamFoundationServerRepositoryBrowser repositoryBrowser) {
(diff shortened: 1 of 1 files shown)


mosabua:
Not sure if I should add any Javadoc for it to explain what it does..

mosabua:
fyi @olivierdagenais  ... this is ready to merge as well. Let me know if you want and doc or changelog or something else.

mosabua:
Btw.. here is the logic 

https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/scm/SCM.java#L146

and the method itself

https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/scm/SCM.java#L752

src/main/java/hudson/plugins/tfs/TeamFoundationServerScm.java
```diff
@@ -414,6 +418,11 @@ public FilePath getModuleRoot(FilePath workspace) {
     public TeamFoundationServerRepositoryBrowser getBrowser() {
         return repositoryBrowser;
     }
+    
+    @Override
+    public @CheckForNull RepositoryBrowser<?> guessBrowser() {
+      return new TeamSystemWebAccessBrowser(serverUrl);
```
olivierdagenais: Would you mind adjusting your IDE's indentation settings for this project to 4 spaces and amending the last commit accordingly?

olivierdagenais:
The integration tests passed.  Approved!  Thank you for your contribution.

PR #86, opened 2016-07-15
Add global configuration of Team Project Collections and associated credentials

olivierdagenais:
These changes add a new section to the global configuration, where one or more pairs of *Team Project Collection* URL and Credentials can be added (and tested!):

image

This also lays the groundwork for [JENKINS-13663: Allow use of credentials from Credentials Plugin](https://issues.jenkins-ci.org/browse/JENKINS-13663).

Manual testing
--------------
As can be seen in the screenshot, I added some credentials to Jenkins and then was able to select them when adding/editing team project collections.

The [Test connection] button mostly works for finding errors in the URL and appears to use NTLM when communicating with TFS (even when credentials are provided), so more testing would be required on that front, especially when the Jenkins master is not in a domain, running as a domain user or on Windows.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -228,6 +228,12 @@
       <artifactId>mailer</artifactId>
       <version>1.16</version>
     </dependency>
+    <dependency>
+      <!-- Used for configuring collections globally -->
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>credentials</artifactId>
+      <version>1.22</version>
+    </dependency>
 
     <dependency>
       <groupId>com.microsoft.tfs.sdk</groupId>
--- a/src/main/java/hudson/plugins/tfs/VstsCollectionConfiguration.java
+++ b/src/main/java/hudson/plugins/tfs/VstsCollectionConfiguration.java
@@ -3,9 +3,13 @@ package hudson.plugins.tfs;
 import hudson.Extension;
 import hudson.model.AbstractDescribableImpl;
 import hudson.model.Descriptor;
+import hudson.util.FormValidation;
 import hudson.util.ListBoxModel;
 import org.kohsuke.stapler.DataBoundConstructor;
 import org.kohsuke.stapler.QueryParameter;
+
+import java.net.MalformedURLException;
+import java.net.URL;
 
 public class VstsCollectionConfiguration extends AbstractDescribableImpl<VstsCollectionConfiguration> {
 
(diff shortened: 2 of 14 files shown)


mosabua:
This looks great. Do you plan to use that global config object to pass around and add things like the showworkspaces flag and so on? 

Also... we might want to add some docs and maybe scripts for how to migrate to this setup as part of the upgrade. E.g. we have hundreds (probably even thousands) of jobs that use the plugin and with the upgrade we would have to bulk change the credentials. We could do that with a groovy script in the script console. Maybe we should write one and make it available as well .. even just as an example ..

olivierdagenais:
> Do you plan to use that global config object to pass around and add things like the showworkspaces flag and so on?

A global "show workspaces" option would certainly belong in `VstsPluginGlobalConfig`, yes.

> Also... we might want to add some docs and maybe scripts for how to migrate to this setup as part of the upgrade. E.g. we have hundreds (probably even thousands) of jobs that use the plugin and with the upgrade we would have to bulk change the credentials. We could do that with a groovy script in the script console. Maybe we should write one and make it available as well .. even just as an example ..

Documentation and automated migration are on the horizon (I've discovered there are extension points we can take advantage of to either have this happen on first run after an upgrade or be user-initiated).  Because nothing currently makes use of this new configuration (plus it would be tricky to get a fully-automated migration right), I didn't think it was necessary to include as part of this pull request.  In fact, it might be wise to slowly introduce something like JENKINS-13663 to prove it out and then once we've had enough brave souls test-drive it for us, we can add user-initiated migration and then upgrade-initiated after that.

aheubusch:
A very welcome feature. +1
Regarding credentials plugin I found this on the mailing list https://groups.google.com/forum/#!searchin/jenkinsci-dev/Upgrading$20Plugin$20to$20use$20Credentials/jenkinsci-dev/Vcrmkqg3oW8/ul6g8Pt4CgAJ
Not very helpful, basically "look at ssh-plugin how they have done it".

A bit of a nitpick. Some commits can be squashed.

olivierdagenais:
So, this is cute and all but as of 2b4ca10, it seems to me this is only decorative and not terribly useful as-is because we are missing some means of referring to a pair of collection URL + credentials.  Having recently been fiddling with the Credentials portion of Jenkins, I'm tempted to add *ID* and *Description* fields like the "Add Credentials" page, pictured below:

image


Also, the [Test connection] button currently only pings TFS whereas I was thinking it would even better to try connecting through the `Server` class or otherwise establish a connection direclty using an instance of `TFSTeamProjectCollection`.

Thoughts?
\ - Oli

PR #88, opened 2016-07-19
Introduce a simple VSTS REST client

olivierdagenais:
In preparation for a fair amount of interactions with VSTS, created the `VstsRestClient` class as a façade.  A sample of its use can be found in `VstsCollectionConfiguration`, for testing the collection URL paired with the credentials.

Manual testing
--------------

1. Temporarily re-enabled the `ping()` test and ran it with the necessary parameters to connect to a VSTS account.  The ping succeeded.
2. Launched a test Jenkins instance and navigated to the global configuration.  For a VSTS collection URL, clicked the [Test connection] button for:
    1. Valid credentials: **Success!**
    2. Invalid credentials: **Error: Server returned HTTP response code: 401 for URL: https://{collectionUrl}/_apis/connectiondata**
    3. No credentials: **Error: VSTS accounts need credentials, preferably a Personal Access Token**

Mission accomplished!

### DIFF
--- a/src/main/java/hudson/plugins/tfs/model/HttpMethod.java
+++ b/src/main/java/hudson/plugins/tfs/model/HttpMethod.java
@@ -1,13 +1,61 @@
 package hudson.plugins.tfs.model;
 
+import org.apache.commons.io.IOUtils;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.net.HttpURLConnection;
+import java.net.ProtocolException;
+import java.nio.charset.Charset;
+
 public enum HttpMethod {
-    GET,
+    GET {
+        @Override
+        public void sendRequest(final HttpURLConnection connection, final String body) throws IOException{
+            // do nothing for GET
+        }
+    },
     POST,
-    HEAD,
+    HEAD {
+        @Override
+        public void sendRequest(final HttpURLConnection connection, final String body) throws IOException{
+            // do nothing for HEAD
+        }
+    },
     PATCH,
     OPTIONS,
     PUT,
     DELETE,
     TRACE,
     ;
+
+    private static final Charset UTF8 = Charset.forName("UTF-8");
+
+    public void sendRequest(final HttpURLConnection connection, final String body) throws IOException {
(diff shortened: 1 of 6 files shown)


src/main/java/hudson/plugins/tfs/model/HttpMethod.java
```diff
@@ -0,0 +1,61 @@
...
+        }
+    },
+    PATCH,
+    OPTIONS,
+    PUT,
+    DELETE,
+    TRACE,
+    ;
+
+    private static final Charset UTF8 = Charset.forName("UTF-8");
```
DavidStaheli: Consider using `java.nio.charset.StandardCharsets.UTF_8.name()` instead.

src/main/java/hudson/plugins/tfs/util/VstsRestClient.java
```diff
@@ -0,0 +1,160 @@
...
+import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URL;
+import java.nio.charset.Charset;
+
+public class VstsRestClient {
+
+    private static final String AUTHORIZATION = "Authorization";
+    private static final String NEW_LINE = System.getProperty("line.separator");
+    private static final Charset UTF8 = Charset.forName("UTF-8");
```
DavidStaheli: Consider using `java.nio.charset.StandardCharsets.UTF_8` instead.

src/main/java/hudson/plugins/tfs/model/HttpMethod.java
olivierdagenais: That's a good idea.  `StandardCharsets` is Java 1.7+ and we're currently targeting Java 1.6, so I'll add an item in the backlog to upgrade to Java 1.7 (as most of Jenkins has) and to make that change at the same time.

DavidStaheli:
Looks good!  See 2 minor comments.

olivierdagenais:
> Looks good! See 2 minor comments.

Thank you for the review!

PR #91, opened 2016-07-23
Report pending and completion status to the associated VSTS commit

olivierdagenais:
This pull request adds a **Set pending status for VSTS commit or pull request** _build step_ as well as a **Set completion status for VSTS commit or pull request** _post-build action_, as shown below in my test job:

image

Manual testing
--------------

1. Triggered the `/vsts-hook/` with a specific commit, which queued my test job.
2. As soon as the Maven build started, I went to check on the commit in VSTS.  It had indeed received the "Pending" status.
3. As soon as the build completed successfully, I went to check on the commit in VSTS.  It had indeed received a "Success" status and was linking back to the Jenkins job.
4. I repeated the test, this time arranging for the build to fail.  VSTS indeed reported a "Failure" status, also linking back to the Jenkins job.

Mission accomplished!

### DIFF
--- a/src/main/java/hudson/plugins/tfs/VstsCompletedStatusPostBuildAction.java
+++ b/src/main/java/hudson/plugins/tfs/VstsCompletedStatusPostBuildAction.java
@@ -1,17 +1,12 @@
 package hudson.plugins.tfs;
 
-import com.cloudbees.plugins.credentials.common.StandardUsernamePasswordCredentials;
 import hudson.Extension;
 import hudson.FilePath;
 import hudson.Launcher;
 import hudson.model.AbstractProject;
-import hudson.model.Result;
 import hudson.model.Run;
 import hudson.model.TaskListener;
-import hudson.plugins.tfs.model.GitCodePushedEventArgs;
-import hudson.plugins.tfs.model.GitStatusState;
-import hudson.plugins.tfs.model.VstsGitStatus;
-import hudson.plugins.tfs.util.VstsRestClient;
+import hudson.plugins.tfs.util.VstsStatus;
 import hudson.tasks.BuildStepDescriptor;
 import hudson.tasks.BuildStepMonitor;
 import hudson.tasks.Notifier;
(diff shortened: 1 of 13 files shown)


src/main/java/hudson/plugins/tfs/VstsCompletedStatusPostBuildAction.java
```diff
@@ -0,0 +1,64 @@
...
+    public static class DescriptorImpl extends BuildStepDescriptor<Publisher> {
+
+        @Override
+        public boolean isApplicable(final Class<? extends AbstractProject> jobType) {
+            return true;
+        }
+
+        @Override
+        public String getDisplayName() {
+            return "Set completion status for VSTS commit or pull request";
```
DavidStaheli: Can we change "VSTS" to "Team Services" here?

src/main/java/hudson/plugins/tfs/VstsPendingStatusBuildStep.java
```diff
@@ -0,0 +1,56 @@
...
+    public static class DescriptorImpl extends BuildStepDescriptor<Builder> {
+
+        @Override
+        public boolean isApplicable(Class<? extends AbstractProject> jobType) {
+            return true;
+        }
+
+        @Override
+        public String getDisplayName() {
+            return "Set pending status for VSTS commit or pull request";
```
DavidStaheli: Can we change "VSTS" to "Team Services" here?

DavidStaheli:
Is there a way for these pre- and post-build actions to be automatic whenever TFS or Team Services is involved, so that administrators do not have to explicitly enable them per-job?

olivierdagenais:
> Is there a way for these pre- and post-build actions to be automatic whenever TFS or Team Services is involved, so that administrators do not have to explicitly enable them per-job?

Yes, I think there’s a way and I’ve added it to the backlog.

src/main/java/hudson/plugins/tfs/VstsPendingStatusBuildStep.java
olivierdagenais: > Can we change "VSTS" to "Team Services" here?

If it's OK with you, I'll make a sweep after merging this and #92.

PR #94, opened 2016-07-27
Clarify the plugin's SCM is for Team Foundation Version Control (TFVC)

olivierdagenais:
One way to prevent JENKINS-33280 is to better document what's expected, since the documentation had fallen behind the implementation details.  These changes provide better examples and explanations of the SCM fields.

Manual testing
--------------

Visit a job configured to use the `TeamFoundationServerScm` and notice only the labels, documentation, etc. have changed and the rest works the same:

image

Mission accomplished!


mosabua:
+1 ... looks good to me. I like all the consistency we are getting ..

olivierdagenais:
Thank you for the review, @mosabua!

PR #97, opened 2016-07-27
Bypass polling when pullRequestMergeCommitCreated event is posted

olivierdagenais:
Upon review of the Jenkins integration specification, it was discovered that there's a need to be able to queue another Jenkins build related to a given pull request if a previous associated build was determined to have failed due to circumstances unrelated to the changes in the pull request, such as a race condition in an unrelated unit test.

Manual testing
--------------

1. Using curl, POST a `gitCodePushed` event and notice that the reply is "Scheduled polling <project>", with no build actually started (because there hasn't been anything new).
2. Using curl, POST a `pullRequestMergeCommitCreated ` event and notice that the reply is "Scheduled <project>", with a corresponding build started, hence bypassing polling.

Mission accomplished!

### DIFF
--- a/src/main/java/hudson/plugins/tfs/TeamPushTrigger.java
+++ b/src/main/java/hudson/plugins/tfs/TeamPushTrigger.java
@@ -40,9 +40,9 @@ public class TeamPushTrigger extends Trigger<Job<?, ?>> {
     public TeamPushTrigger() {
     }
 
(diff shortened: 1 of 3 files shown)


DavidStaheli:
Looks good.  Thanks, Oli!

olivierdagenais:
Thank you for the code review!

PR #99, opened 2016-07-29
Preliminary team-build endpoint

olivierdagenais:
With these changes, we've got a barely-working `/team-build/build/JOB_NAME` implementation that works a bit like `/job/JOB_NAME/build`, only _without_ token support and ONLY for the `PUT` method.

Notice the `JOB_NAME` is still in the URL, just at a different spot.  Now that I actually write it out like that, it seems kind of silly and I _might_ be able to make it like `/team-build/JOB_NAME/build` instead.

Manual testing
--------------

1. Launch local Jenkins development instance
2. Activate security
    1. Create an account with user ID `remote`
    2. Sign in to Jenkins as `remote`
    3. Go to http://localhost:9090/me/configure and record the API token.  (I'll be using `92e9d8998a8c005697d252f09a2a311b` below)
3. Create the `ping.json` file with the following content:
    ```json
    {
        "message": "Hello, world!"
    }
    ```

4. Run `curl --request PUT http://localhost:9090/team-build/build/Git --data-urlencode json@ping.json --user remote:92e9d8998a8c005697d252f09a2a311b` and:
    1. Notice that the console output says a queue item was created.
    2. Notice that a build of the `Git` project was queued.
    3. Go take a look at the revision that was built.  For me it was the latest at the remote's origin/master.
5. Create the `build.json` file with the following content:
    ```json
    {
        "team-parameters":
        {
            "collectionUri":"https://mseng.visualstudio.com",
            "repoUri":"https://mseng.visualstudio.com/Personal/_git/olivida.gcm4ml",
            "projectId":"Personal",
            "repoId":"olivida.gcm4ml",
            "commit":"6a23fc7afec31f0a14bade6544bed4f16492e6d2",
            "pushedBy":"olivida"
        }
    }
    ```

6. Run `curl --request PUT http://localhost:9090/team-build/build/Git --data-urlencode json@build.json --user remote:92e9d8998a8c005697d252f09a2a311b` and:
    1. Notice that the console output says a queue item was created.
    2. Notice that a build of the `Git` project was queued.
    3. Go take a look at the revision that was built.  That's right, it was `6a23fc7afec31f0a14bade6544bed4f16492e6d2`.
    4. Go to Team Services and notice that commit `6a23fc7afec31f0a14bade6544bed4f16492e6d2` has had new statuses posted as a result of the build we just queued.  BOOYEAH!

Mission accomplished!


src/main/java/hudson/plugins/tfs/TeamBuildEndpoint.java
```diff
@@ -0,0 +1,223 @@
...
+
+    public String getJobName() {
+        return jobName;
+    }
+
+    boolean decodeCommandAndJobNames(final String pathInfo) {
+        if (pathInfo.startsWith(URL_PREFIX)) {
+            final String restOfPath = pathInfo.substring(URL_PREFIX.length());
+            final int firstSlash = restOfPath.indexOf('/');
+            if (firstSlash != -1) {
```
AlexRukhlin: What if firstSlash == 0?

src/main/java/hudson/plugins/tfs/TeamBuildEndpoint.java
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    public String getJobName() {
+        return jobName;
+    }
+
+    boolean decodeCommandAndJobNames(final String pathInfo) {
+        if (pathInfo.startsWith(URL_PREFIX)) {
+            final String restOfPath = pathInfo.substring(URL_PREFIX.length());
+            final int firstSlash = restOfPath.indexOf('/');
```
AlexRukhlin: Note that the job might be nested in some folders, and the URL could look like:
http://tfsxp-build:9090/job/Folder1/job/Folder2/job/AR_Test_02/buildWithParameters

src/main/java/hudson/plugins/tfs/TeamBuildEndpoint.java
```diff
@@ -0,0 +1,223 @@
...
+            return;    // everyone is authorized
+        }
+
+        project.checkPermission(Item.BUILD);
+
+        if (req.getMethod().equals("PUT")) {
+            return;
+        }
+
+        throw HttpResponses.error(SC_METHOD_NOT_ALLOWED, "Only PUT is supported");
```
AlexRukhlin: @olivierdagenais What is the reason for this?

src/main/java/hudson/plugins/tfs/model/PingCommand.java
```diff
@@ -0,0 +1,61 @@
...
+
+        @Override
+        public String getSampleRequestPayload() {
+            return "{\n" +
+                    "    \"parameter\":\n" +
+                    "    [\n" +
+                    "        {\"name\":\"id\",\"value\":\"123\"},\n" +
+                    "        {\"name\":\"verbosity\",\"value\":\"high\"}\n" +
+                    "    ],\n" +
+                    "    \"team-parameters\":\n" +
```
AlexRukhlin: Shouldn't it be "team-parameter" (i.e. singular) to be consistent with the "parameter" item?

src/main/java/hudson/plugins/tfs/TeamBuildEndpoint.java
olivierdagenais: > What if firstSlash == 0?

Interesting scenario.  `firstSlash` will be `0` happens when the input looks like `/team-build//ping/a` and therefore the `commandName` is decoded as the empty string, while the `jobName` is decoded as `ping/a`.  

The caller (`dispatch()`) responds with HTTP 400 (Bad request) saying "Command not implemented" as a result.

src/main/java/hudson/plugins/tfs/model/PingCommand.java
olivierdagenais: Tempting, but I'd rather pluralize ours to be consistent with its content.

olivierdagenais:
Thank you for your code review comments!

PR #100, opened 2016-08-02
Support non-specific pull request status

olivierdagenais:
It turns out we don't always know which pull request iteration a build might be requested for, so added a method for that REST endpoint and now calling it accordingly.

Manual testing
--------------

1. Using curl, POST a `pullRequestMergeCommitCreated` event as before and notice that the corresponding pull request iteration status is still reported as before this change.
2. Temporarily remove `iterationId` from the payload, repeat the above test and notice that the corresponding pull request status (without an iterationId) is now reported.

Mission accomplished!


DavidStaheli:
Looks good

olivierdagenais:
Thank you for your code review!

PR #103, opened 2016-08-03
Make permission check behave exactly like Jenkins'

olivierdagenais:
This adds support for the "Trigger builds remotely (e.g., from scripts)" trigger (which has an authentication token), as well as POST with crumbs, when CSRF protection is enabled.

Manual testing
--------------

1. Enabled "Prevent Cross Site Request Forgery exploits" in the Global Security.
2. Using `curl`, fetched the crumb for a user, then inserted it as a header when I POSTed to the `/team-build/build/` endpoint for a job.  A build was queued.
3. Created a job that used the "Trigger builds remotely (e.g., from scripts)" trigger, then used `curl` to PUT to the `/team-build/build/` endpoint for it, providing the `token` as a form body parameter.  A build was queued.

Mission accomplished!


jasholl:
looks good to me.

PR #106, opened 2016-08-04
Deactivate logging of each resource get in build log

mosabua:
Use the same approach as in the list workspaces command.

- wrapped flag around logging
- set flag to false by default
- added comment to add property at later stage if desired
- updated unit test

Also added count of resources downloaded and added that to the log.

Verified in dev instance of Jenkins with numerous validations jobs. Also should pass cloudbees build.


mosabua:
Would be good to get this into the release @olivierdagenais @DavidStaheli

mosabua:
Example output from a test job:

Getting version 'C3936' to '/build/workspace/admin/Verify/tfs_dev_sampleApp_build'...
Finished getting version 'C3936'. Retrieved 311 resources.

olivierdagenais:
Nice and simple, approved!

Thank you for your contribution!

mosabua:
Awesome .. you rock @olivierdagenais

PR #110, opened 2016-08-10
Reduce the chance of error with Team Services URLs

olivierdagenais:
TFS eventually acquired the concept of Team Project Collections to allow isolation between groups/projects (on the same server) and Visual Studio Team Services initially inherited this functionality by requiring `/DefaultCollection` in the URL.  After a while, Team Services removed the requirement of specifying `/DefaultCollection` in the URL while still supporting URLs with it.  This lead to a situation where the plugin would consider two functionally-identical URLs as different collections.

This pull request aims to mitigate the confusion by:

1. Enforcing the lack of a path in Team Services collection URLs in the global configuration.
2. Emitting warnings when jobs _almost match_ a given event URL.
3. Emitting warnings when no jobs exactly matched a given event URL.

Manual testing
--------------
The test cases are in the same order as the mitigations above.

1. Attempted to add `/DefaultCollection` to a Team Services collection URL: image (no such error was shown when the URL's host name did not end with `.visualstudio.com`)
2. Purposely added a `/DefaultCollection` to the value of the `remoteUrl` in an event payload of type `git.pullrequest.merged` and then POSTed it to the `/team-events/gitPullRequestMerged` endpoint.  My jobs that were configured with a remote URL equivalent to the same repository (minus the `/DefaultCollection`) were logged as _almost matched_ in a warning and were not triggered.
3. Since none of the jobs had `/DefaultCollection` in a Git remote URL, there was a warning saying exactly that.

Mission accomplished!

### DIFF
--- a/src/main/java/hudson/plugins/tfs/TeamCollectionConfiguration.java
+++ b/src/main/java/hudson/plugins/tfs/TeamCollectionConfiguration.java
@@ -70,6 +70,10 @@ public class TeamCollectionConfiguration extends AbstractDescribableImpl<TeamCol
                 return FormValidation.error("Malformed TFS/Team Services collection URL (%s)", e.getMessage());
             }
 
+            final String hostName = uri.getHost();
+            if (isTeamServices(hostName)) {
+                return checkTeamServices(uri);
+            }
             // TODO: check that it's not a deep URL to a repository, work item, API endpoint, etc.
 
             return FormValidation.ok();
@@ -131,6 +135,16 @@ public class TeamCollectionConfiguration extends AbstractDescribableImpl<TeamCol
                     .withEmptySelection()
                     .withAll(matches);
         }
+    }
+
+    static FormValidation checkTeamServices(final URI uri) {
+        final String path = uri.getPath();
+        if (path != null) {
+            if (path.length() > 0 && !path.equals("/")) {
+                return FormValidation.error("A Team Services collection URL must have an empty path.");
+            }
+        }
+        return FormValidation.ok();
     }
 
     static boolean isTeamServices(final String hostName) {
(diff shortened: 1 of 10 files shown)


DavidStaheli:
This code looks good, and thanks for the tests.  The UI is confusing to me though.  It asks for a "Collection URL" which I interpret as "the URL that includes `/DefaultCollection` on its path."  Instead of prohibiting this form of URL and displaying an error, could we just treat `abc.visualstudio.com/DefaultCollection` as `abc.visualstudio.com` behind the scenes?

olivierdagenais:
> The UI is confusing to me though. It asks for a "Collection URL" which I interpret as "the URL that includes /DefaultCollection on its path."

Yes, historically, for Team Services, that part was required.  I thought it was now deprecated.

> Instead of prohibiting this form of URL and displaying an error, could we just treat abc.visualstudio.com/DefaultCollection as abc.visualstudio.com behind the scenes?

I might be able to re-write the value in the form such that the next time you reload it, it would be missing the path if it was a Team Services account.

I tried fuzzy-matching URLs with and without `/DefaultCollection` during the events' enumeration of jobs (using the `isTeamServicesNearMatch()` method) and although I was able to find the jobs and trigger them, the Git plugin still uses a strict check and would not build the requested commit because none of the remote URLs matched what was provided.  Rather than having something look like it's working and doing the wrong thing, I opted to not trigger the jobs and instead emit a warning saying there was an "almost match".

Do you think the warning message should be clearer about the fact that there's a configuration mismatch/error?  Currently, users would see something like:

`WARNING: Remote URL 'https://fabrikam-fiber-inc.visualstudio.com/DefaultCollection/Fabrikam/_git/FabrikamGit' of job 'FabrikamGit-CI' almost matched event URL 'https://fabrikam-fiber-inc.visualstudio.com/Fabrikam/_git/FabrikamGit'.`

Ideally, I would like to detect, at configuration time, when a remote URL used in the Git plugin's SCM section refers to a Git repository in Team Services that has an extra `/DefaultCollection` in the URL, however it did not look like I had that ability from my plugin.  We could always contribute such a check to the Git plugin such that new versions would have it, but I'm not sure if that makes sense.

PR #116, opened 2016-08-12
Switch to Jackson for events parsing

olivierdagenais:
This branch contains no [direct] new features, just some refactoring, re-wiring and clean-up that will make it easier to add upcoming features.


mosabua:
I like the switch to Jackson. Its a great library. You should probably add an explicit dependency to it in the pom.

olivierdagenais:
> You should probably add an explicit dependency to it in the pom.

I'm using the version that ships with the TFS SDK for Java (it's in the JAR) and thus I'm weary of over-specifying dependencies.

mosabua:
Oh.. I wasn't aware that Jackson is repackaged there. I guess thats okay then... bad practice on the TFS side, but nothing we can control.

PR #119, opened 2016-08-16
Add support for per-computer TFVC configuration folders

olivierdagenais:
In some environments, the "home" directory is mounted over a network and shared between many computers, including Jenkins servers and their associated build nodes, which eventually leads to corruption of the configuration directory used for TFVC workspaces.  This branch introduces a new, advanced setting that uses a sub-directory for each computer, copying the current configuration directory (if one is found) on the next run.

In other words, turning on the new **Store TFVC configuration in computer-specific folders** option means the TFS SDK will use the following base path for its <code>Cache</code>, <code>Configuration</code> and <code>Logs</code> directories:
`~/.microsoft/Team Foundation/4.0/${HOSTNAME}/`

Manual testing
--------------

1. **Jenkins** > **Manage Jenkins** > **Configure System** and scroll to **TFS/Team Services**
2. Check the new **Store TFVC configuration in computer-specific folders** checkbox and click **Save**:
image
3. Queue a job on MASTER that uses TFVC for its SCM.
    1. Notice that the `4.0` folder now contains a sub-folder named after the server's host name, with copies of the existing folders.
4. Modify the job to execute on a remote node and queue it again.
    1. Notice that the `4.0` folder on the _node_ now contains a sub-folder named after the node's host name.

Mission accomplished!

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ To make use of the Git integration with TFS/Team Services, it is necessary to fi
 
 ### Advanced
 
-In some environments, the "home" directory is mounted over a network and shared between many build nodes, which eventually leads to corruption of the configuration directory used for TFVC workspaces. If you have such an environment, check the box next to **Store TFVC configuration in node-specific folders** to use a sub-directory for each node. :warning: WARNING :warning: Turning this on is equivalent to setting the `TEE_PROFILE_DIRECTORY` environment variable and thus any manual operations performed on nodes using the Command-Line Client (CLC) will need to be performed with the `TEE_PROFILE_DIRECTORY` environment variable set accordingly.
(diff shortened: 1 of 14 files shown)


olivierdagenais:
There's another case to handle: more than one Jenkins server sharing the same home folder.  It has been suggested to scope by Jenkins server hostname in that case.  For example, this is what the base folder could look like:

1. `.../4.0/jenkins-1/MASTER/`
1. `.../4.0/jenkins-1/MASTER/`
1. `.../4.0/jenkins-1/MASTER/`

olivierdagenais:
> It has been suggested to scope by Jenkins server hostname in that case.

This turned out to be a suboptimal idea and the host's "short name" was used, instead.  It solves the multi-master problem, plus a few more.  See 2231f6121fec20de65e27594567d397a784b800d for more details.

olivierdagenais:
In an offline discussion, the label for the new checkbox came up as being unclear.  Here are some ideas for alternative wording:

* My home folder is a shared resource
* Support shared home folders by using sub-folders
* Prevent shared configuration conflicts
* Relocate TEE persistence store

PR #129, opened 2016-09-01
Fix error when building URL using UriHelper#join

olivierdagenais:


DavidStaheli:
Looks good.  Thanks, Oli!

mosabua:
Ouch.. people actually use repo names with spaces in them.. sigh? TFS should disallow that on the server imho.. but its too late for that ;-)

olivierdagenais:
> people actually use repo names with spaces in them

It can also be Team Project names, build definition names, etc.

PR #130, opened 2016-09-01
Converted ERROR into NOTICE & improved wording

olivierdagenais:
Manual testing
--------------

1. Configured a job with the **Set build pending status in TFS/Team Services** build step and the **Set build completion status in TFS/Team Services** post-build action.
2. Triggered a build via the `/team-build/` endpoint with a payload containing a `team-build` section (just like the Jenkins Queue Job task would do).
    1. The _Console Output_ contained: `NOTICE: You selected 'Set build pending status in TFS/Team Services' on your Jenkins job, but this option has no effect when calling the job from the 'Jenkins Queue Job' task in TFS/Team Services.` as well as `NOTICE: You selected 'Set build completion status in TFS/Team Services' on your Jenkins job, but this option has no effect when calling the job from the 'Jenkins Queue Job' task in TFS/Team Services.`
    2. The Jenkins build still linked to the corresponding Team Services build.

Mission accomplished!

### DIFF
--- a/tfs/src/main/java/hudson/plugins/tfs/TeamCompletedStatusPostBuildAction.java
+++ b/tfs/src/main/java/hudson/plugins/tfs/TeamCompletedStatusPostBuildAction.java
@@ -4,11 +4,13 @@ import hudson.Extension;
 import hudson.FilePath;
 import hudson.Launcher;
 import hudson.model.AbstractProject;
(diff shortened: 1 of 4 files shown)


jasholl:
Looks good to me

DavidStaheli:
+1

PR #131, opened 2016-09-02
Collect build results for an upstream TFS/Team Services build

olivierdagenais:
This adds the **Collect results for TFS/Team Services** post-build action, where we can configure a list of results to collect, each with their own "includes":

image

Manual testing
--------------

1. Configured a Maven job with the new post-build action, as per the screenshot above.
2. Queued a build of the job.
3. Near the end of the build, the console contained: `Recording results... JUnit (13 files) Cobertura (1 file). Compressing... Done.`
4. Once the build completed, visiting **/job/JOB_NAME/BUILD_NUMBER/team-results/zip** downloaded the file `zip.zip`, which contained the following structure:

    ```
    team-results
        cobertura
            reports
                jacoco.exec
        junit
            target
                surefire-reports
                    TEST-com.microsoft.alm.authentication.AzureAuthorityTest.xml
                    TEST-com.microsoft.alm.authentication.AzureDeviceFlowResponseTest.xml
                    (...)
    ```

Mission accomplished!

### DIFF
--- a/tfs/src/main/java/hudson/plugins/tfs/model/TeamResultType.java
+++ b/tfs/src/main/java/hudson/plugins/tfs/model/TeamResultType.java
@@ -1,0 +1,36 @@
+package hudson.plugins.tfs.model;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+public enum TeamResultType {
+    JUNIT("junit", "JUnit", Collections.singletonList("**/TEST-*.xml")),
+    MAVEN("junit", "Maven", Arrays.asList(
+            "**/surefire-reports/TEST-*.xml",
+            "**/failsafe-reports/TEST-*.xml"
+    )),
+    ;
+
+    private final String folderName;
+    private final String displayName;
+    private final List<String> defaultPatterns;
+
+    TeamResultType(final String folderName, final String displayName, final List<String> defaultPatterns) {
+        this.folderName = folderName;
+        this.displayName = displayName;
+        this.defaultPatterns = defaultPatterns;
+    }
+
+    public String getFolderName() {
+        return folderName;
+    }
+
+    public String getDisplayName() {
+        return displayName;
+    }
+
+    public List<String> getDefaultPatterns() {
+        return defaultPatterns;
+    }
+}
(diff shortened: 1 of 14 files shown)


DavidStaheli:
The TeamResultType "Maven" might be confusing.  Selecting "Maven" as the type of "Build results to collect" doesn't correlate to test results in my mind.  What do you think about naming it "Maven JUnit" or removing it altogether since the "JUnit" type should suffice?

DavidStaheli:
Selecting a single TeamResultType conflicts with when multiple result types need to be collected.  For example, a user may want to collect JUnit and Cobertura results.  Can we avoid selecting a type altogether and let users specify a custom mix of glob patterns?

olivierdagenais:
> The TeamResultType "Maven" might be confusing.

Yeah, it was an idea I had to collect what are, as you pointed out, JUnit result files with a special glob pattern.  Indeed, if you select "Maven", those files end up in a `junit` folder.

DavidStaheli:
Ah, does the UI allow adding multiple TeamResultTypes to collect?

olivierdagenais:
> Ah, does the UI allow adding multiple TeamResultTypes to collect?

Yes, that's what the [Add] button does in the first screenshot.  I should have shown more than one.

tfs/src/main/java/hudson/plugins/tfs/model/TeamResultType.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+public enum TeamResultType {
+    JUNIT("junit", "JUnit", Collections.singletonList("**/TEST-*.xml")),
+    MAVEN("junit", "Maven", Arrays.asList(
+            "**/surefire-reports/TEST-*.xml",
+            "**/failsafe-reports/TEST-*.xml"
+    )),
```
olivierdagenais: I just put these two as placeholders and I am hoping this pull request will serve as a conversation starter to decide what result types we'll want to add here.

olivierdagenais:
> The TeamResultType "Maven" might be confusing.

Removed!  See the new list based on your suggestion.

DavidStaheli:
Looks good!  Thanks, Oli.

PR #134, opened 2016-09-07
Automatic integration

olivierdagenais:
This pull request adds two "global options" to Jenkins, pictured below:

1. **Enable Push Trigger for all jobs**
2. **Enable Team Status for all jobs**

image

Manual testing
==============

1. Enable both new options and configure a few jobs with special cases:
    1. One job to opt-out of the global push trigger.
    2. One job with the "Build when a change is pushed to TFS/Team Services" trigger enabled.
    3. One job with the "Set build pending status in TFS/Team Services" build step.
    4. One job with the "Set build completion status in TFS/Team Services" post-build action.
    5. One job with both the "Set build pending status in TFS/Team Services" build step and the "Set build completion status in TFS/Team Services" post-build action.
2. Hit the `/team-events/gitPullRequestMerged` endpoint with a payload that matches the Git repository used by all the jobs above.
    1. All the jobs that used the Git SCM with the same repository URL as the event, EXCEPT the job that had opted out, were queued.
    2. Looking at the statuses posted to the Git commit and the pull request in Team Services, there were both a pending status and a completed status for each build that ran, regardless of overlap in configuration.
3. Disabled both new options and hit the same endpoint with the same payload.
    1. Only the job that had the push trigger enabled was triggered.
    2. It only posted the completion status to the commit and pull request (that was how it was configured).

Mission accomplished!

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,8 +31,11 @@ The plugin will automatically create a workspace in TFS/Team Services and map a 
 
 The TFS plug-in for Jenkins enhances the Git plug-in for Jenkins by adding some integration features:
 * A push trigger, to request builds of specific commits in Git repositories without needing to schedule SCM polling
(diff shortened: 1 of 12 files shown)


DavidStaheli:
Looks good.  Thanks, Oli!

PR #136, opened 2016-09-12
Fuzzy TFS/Team Services URL matching

olivierdagenais:
As a follow-up to #110, we've since discovered a way to support TFS/Team Services Git repository URLs with _and_ without `/DefaultCollection` (see ebdee65b7dde940a1ecf3e2c88962ad64bf6ecef for details).

Manual testing
==============

1. Configure three Jenkins jobs with the Git SCM, pointing to three variations of effectively the same Team Services Git repository URL, making sure to add the **Set build completion status in TFS/Team Services** post-build action:
    1. HTTPS, _with_ `/DefaultCollection`.  For example: `https://fabrikam-fiber-inc.visualstudio.com/DefaultCollection/project/_git/repo`
    1. HTTPS, _without_ `/DefaultCollection`.  For example: `https://fabrikam-fiber-inc.visualstudio.com/project/_git/repo`
    1. SSH, _with_ `/DefaultCollection`.  For example: `ssh://fabrikam-fiber-inc@fabrikam-fiber-inc.visualstudio.com/DefaultCollection/project/_git/repo`
    1. SSH, _without_ `/DefaultCollection`.  For example: `ssh://fabrikam-fiber-inc@fabrikam-fiber-inc.visualstudio.com/project/_git/repo`
2. Manually queue each job once to make sure it's able to fetch from Git using the configured credentials, etc.
3. Trigger a "Git pull request merged" event that contains a Git repository remote URL _without_ `/DefaultCollection`.  For example: `https://fabrikam-fiber-inc.visualstudio.com/project/_git/repo`
    1. Notice that all four jobs were queued.
    2. Notice that the Git plugin reports building the same commit for all four jobs (the commit ID came from the event payload).
    3. Notice that the list of statuses for that commit, in Team Services, includes links back to each of the jobs with the completion status.
4. Trigger a "Git pull request merged" event with URLs _containing_ `/DefaultCollection` in the payload.  For example: `https://fabrikam-fiber-inc.visualstudio.com/DefaultCollection/project/_git/repo`
    1. Notice that all four jobs were queued.
    2. Notice that the Git plugin reports building the same commit for all four jobs (the commit ID came from the event payload).
    3. Notice that the list of statuses for that commit, in Team Services, includes links back to each of the jobs with the completion status.

Mission accomplished!

PR #140, opened 2016-11-22
User account mapping

smalik86:
Hi @olivierdagenais , these are my testing results.

I ran the build through Maven and made sure that the jar and hpi files were successfully created.
Once the hpi file was created, I uploaded the file and installed it on our company's Test Jenkins.
The plugin worked as expected and the UI interface made it easy for us to chose whether to keep the User ID with the domain name or User ID without the domain name. 

Example: DomainName/User ID to Only User ID
tfs ui

### DIFF
--- a/tfs/src/main/java/hudson/plugins/tfs/TeamPluginGlobalConfig.java
+++ b/tfs/src/main/java/hudson/plugins/tfs/TeamPluginGlobalConfig.java
@@ -2,6 +2,9 @@ package hudson.plugins.tfs;
 
 import hudson.Extension;
 import hudson.ExtensionList;
+import hudson.plugins.tfs.model.DomainUserAccountMapper;
+import hudson.plugins.tfs.model.UserAccountMapper;
+import hudson.plugins.tfs.model.UserAccountMapperDescriptor;
 import jenkins.model.GlobalConfiguration;
 import jenkins.model.Jenkins;
 import net.sf.json.JSONObject;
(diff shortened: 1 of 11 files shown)


olivierdagenais:
Those are some mice commits, @smalik86! 😃

Given I haven't personally done any testing, I'm assuming you found these changes to be satisfactory at your organization?  If so, can I trouble you to add a comment to this pull request with the validation you performed?  A great example of what I'm looking for can be seen in the **Manual Testing** section of #136.

Thanks!
\- Oli

smoyen:
# Manual testing
Test 1: When **User account name mapping strategy** is set to _Resolve user using ‘alias’ only, removing ‘DOMAIN\’_
1.	Create a job using TFS for source control and turn polling on 
2.	Check in a change
3.	After the build kicks off, examine the user listed in the changes. Notice that it **does not** include the domain 
4.	If your Jenkins user ids **do not** include the domain, you will have the added benefit of the build history for a user including build kicked off Manually and by SCM Trigger


Test 2: When **User account name mapping strategy** is set to _Resolve user using ‘DOMAIN\alias’_
1.	Create a job using TFS for source control and turn polling on 
2.	Check in a change
3.	After the build kicks off, examine the user listed in the changes. Notice that it **does** include the domain
4.	If your Jenkins user ids **do not** include the domain, the build histories for builds kicked off Manually and by SCM Trigger will be listed under separate users (i.e. Manually builds will be listed under user named _userId_ and SCM Trigger builds will be listed under user named _domainName_userId_) 

*Jobs that use anything besides TFS for source control are not affected. Manual Build user name is not affected.

olivierdagenais:
@smalik86 and @smoyen: Thank you for validating these changes!  Your contribution is greatly appreciated.

PR #141, opened 2016-12-01
getCommits can return null

CSchulz:
```
Dez 01, 2016 1:47:22 PM SCHWERWIEGEND hudson.plugins.tfs.TeamEventsEndpoint dispatch
Error while performing reaction to 'gitPush' event.
java.lang.NullPointerException
  at hudson.plugins.tfs.model.GitPushEvent.determineCommit(GitPushEvent.java:100)
  at hudson.plugins.tfs.model.GitPushEvent.decodeGitPush(GitPushEvent.java:120)
  at hudson.plugins.tfs.model.GitPushEvent.perform(GitPushEvent.java:44)
  at hudson.plugins.tfs.TeamEventsEndpoint.innerDispatch(TeamEventsEndpoint.java:163)
  at hudson.plugins.tfs.TeamEventsEndpoint.dispatch(TeamEventsEndpoint.java:133)
  at hudson.plugins.tfs.TeamEventsEndpoint.doGitPush(TeamEventsEndpoint.java:201) 
```

### DIFF
--- a/tfs/src/main/java/hudson/plugins/tfs/model/AbstractHookEvent.java
+++ b/tfs/src/main/java/hudson/plugins/tfs/model/AbstractHookEvent.java
@@ -85,6 +85,10 @@ public abstract class AbstractHookEvent {
     List<GitStatus.ResponseContributor> pollOrQueueFromEvent(final GitCodePushedEventArgs gitCodePushedEventArgs, final List<Action> actions, final boolean bypassPolling) {
         List<GitStatus.ResponseContributor> result = new ArrayList<GitStatus.ResponseContributor>();
         final String commit = gitCodePushedEventArgs.commit;
+        if (commit == null) {
+            result.add(new GitStatus.MessageResponseContributor("No commits were pushed, skipping further event processing."));
+            return result;
+        }
         final URIish uri = gitCodePushedEventArgs.getRepoURIish();
 
         TeamGlobalStatusAction.addIfApplicable(actions);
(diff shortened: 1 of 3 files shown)


cniweb:
This will not fix the problem.
Commits can be empty:
https://www.visualstudio.com/en-us/docs/integrate/get-started/service-hooks/events#git.push
But why?

Chris

DavidStaheli:
@cniweb Unfortunately, when a simple tag/branch is pushed and contains no commits, `gitPush.getCommits()` will return null instead of an empty list.  In my opinion, this pull request has the right approach:  `if (commits == null || commits.size() < 1)`

olivierdagenais:
I think I agree with @cniweb: this change replaces `NullPointerException` with `IllegalArgumentException`, meaning the hook would return HTTP 400 instead of HTTP 500.

It seems to me it might be best to skip the `gitPush` event's reaction (i.e. polling) if no commits were actually pushed, possibly also emitting an informational message to the log.

This would be implemented by `GitPushEvent#determineCommit` returning `null` if it can't find a commit and then `AbstractHookEvent#pollOrQueueFromEvent` checking if `commit` is `null` and adding an appropriate response to `result` if that's the case, kind of like the one that adds `No Git jobs found`.

What do you think, @CSchulz  and @DavidStaheli?

Thanks!
\- Oli

olivierdagenais:
@CSchulz: I have added commit cba1ab01055d048e432d5848b835223e39d558d5 (which implements my above suggestion) to your branch and pushed it, effectively updating this pull request.

What do you think?

Thanks!
\- Oli

olivierdagenais:
Thank you all for your help on this matter!

CSchulz:
Nice idea, I was on vacation. :sunglasses: 

What about rebase (force) pushes? I think they won't get processed, right?

cniweb:
Hi @olivierdagenais,

When will a new release be released?

Chris

olivierdagenais:
> When will a new release be released?

[5.3.1](https://github.com/jenkinsci/tfs-plugin/releases/tag/tfs-5.3.1) was released and should show up soon in the update center.

Thank you for your patience!

\- Oli

PR #142, opened 2016-12-07
Fix exception when invoked from TFS Release Management (JENKINS-40283)

DavidStaheli:
This addresses issue JENKINS-40283 that addresses a crash with the following stack trace:

```
java.lang.Error: java.net.URISyntaxException: Cannot parse Git URI-ish: The uri was empty or null
	at hudson.plugins.tfs.model.GitCodePushedEventArgs.getRepoURIish(GitCodePushedEventArgs.java:23)
	at hudson.plugins.tfs.CommitParameterAction.&lt;init>(CommitParameterAction.java:20)
	at hudson.plugins.tfs.model.BuildCommand.contributeTeamBuildParameterActions(BuildCommand.java:227)
	at hudson.plugins.tfs.model.BuildCommand.perform(BuildCommand.java:110)
```

**Tested as follows:**
1. `mvn clean verify` - All tests pass
2. I verified failure with the above stack trace when the "Jenkins Queue Job" task invoked a Jenkins job from TFS Release Management.  Then, I applied this code change and verified that the job ran successfully.

### DIFF
--- a/tfs/src/main/java/hudson/plugins/tfs/model/BuildCommand.java
+++ b/tfs/src/main/java/hudson/plugins/tfs/model/BuildCommand.java
@@ -29,6 +29,7 @@ import jenkins.util.TimeDuration;
 import net.sf.json.JSONArray;
 import net.sf.json.JSONObject;
 import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang.StringUtils;
 import org.kohsuke.stapler.StaplerRequest;
 
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


AlexRukhlin:
LGTM

tfs/src/main/java/hudson/plugins/tfs/model/BuildCommand.java
```diff
@@ -209,23 +209,27 @@ static void contributeTeamBuildParameterActions(final Map<String, String> teamBu
...
-                final GitCodePushedEventArgs args = new GitCodePushedEventArgs();
-                args.collectionUri = collectionUri;
-                args.repoUri = repoUri;
-                args.projectId = projectId;
-                args.repoId = repoId;
-                args.commit = commit;
-                args.pushedBy = pushedBy;
-                final CommitParameterAction action = new CommitParameterAction(args);
-                actions.add(action);
+                if (repoUriString != null && !repoUriString.trim().isEmpty()) {
```
olivierdagenais: Is `Build.Repository.Uri` present in the JSON payload and has an empty/null value or is it not present at all?  For example, [line 207](https://github.com/jenkinsci/tfs-plugin/pull/142/commits/465021e4c1bcbab09fb61022c2fd539515771988#diff-f7db47497a32ee414942dd7079038aa5R207) above checks if the map contains an entry with what resolves to the `Build.Repository.Provider` key.

Otherwise, might I recommend using `org.apache.commons.lang.StringUtils.isNotBlank()`?

DavidStaheli: `Build.Repository.Uri` is present in the JSON payload and ends up having a value of white space.  So unfortunately, we can't use `teamBuildParameters.containsKey(BUILD_REPOSITORY_URI)`.  Great suggestion to use `org.apache.commons.lang.StringUtils.isNotBlank()`.  I made that change.  Does it look okay otherwise @olivierdagenais?

PR #147, opened 2017-01-27
Upgrade to TFS SDK 14.0.4

nexaddo:


nexaddo:
I guessing Microsoft removed some things from the TFS SDK and that's the reason the SDK hasn't been updated past 14.0.3

PR #148, opened 2017-02-08
fix memory leak: Workstation.currentWorkstation

holgercn:
if isConfigFolderPerNode enabled, every checkout create a new ClonePersistenceStoreProvider instance in
Workstation.currentWorkstation HashMap and never be deleted


mosabua:
Lgtm 👍

PR #149, opened 2017-02-15
Support WorkflowJob (Pipeline) jobs

yacaovsnc:
Today when triggering a Pipeline job, we will get a "Project not found" exception as `WorkflowJob` is not of `AbstractProject` type.  

All information the plugin requires can be retrieved by `BuildableItem` and `ParameterizedJobMixin.ParameterizedJob` interface, and `WorkflowJob` implements both interfaces.  The change is to explicitly cast both `AbstractProject` and `WorkflowJob` down to `BuildableItem` and `ParameterizedJobMixin.ParameterizedJob` and pass it around as such.

### DIFF
--- a/tfs/src/main/java/hudson/plugins/tfs/TeamBuildEndpoint.java
+++ b/tfs/src/main/java/hudson/plugins/tfs/TeamBuildEndpoint.java
@@ -2,8 +2,9 @@ package hudson.plugins.tfs;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import hudson.Extension;
-import hudson.model.BuildableItem;
 import hudson.model.BuildAuthorizationToken;
+import hudson.model.BuildableItem;
+import hudson.model.Item;
 import hudson.model.Job;
 import hudson.model.UnprotectedRootAction;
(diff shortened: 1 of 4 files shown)


tfs/src/main/java/hudson/plugins/tfs/TeamBuildEndpoint.java
```diff
@@ -54,6 +56,7 @@
     private static final Map<String, AbstractCommand.Factory> COMMAND_FACTORIES_BY_NAME;
     public static final String URL_NAME = "team-build";
     public static final String PARAMETER = "parameter";
+    public static final String BUILD_SOURCE_BRANCH = "Build.SourceBranch";
```
DavidStaheli: Do you know if `Build.SourceBranch` is available in VSTS Release Management scenarios?  If not, can we handle its absence gracefully?

yacaovsnc: Great call.  Made changes to get the branch from:
1. jobName input.
2. Build.SourceBranch variable

Throw if none of them are available.

PR #152, opened 2017-03-02
Overwrite when getting files

pescuma:
To avoid problems when files are changed during the build.

### DIFF
--- a/tfs/src/main/java/hudson/plugins/tfs/TeamFoundationServerScm.java
+++ b/tfs/src/main/java/hudson/plugins/tfs/TeamFoundationServerScm.java
@@ -91,7 +91,8 @@ public class TeamFoundationServerScm extends SCM {
     private String userName;
     private CredentialsConfigurer credentialsConfigurer;
     private boolean useUpdate;
-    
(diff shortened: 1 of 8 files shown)


mosabua:
Should this maybe be an option ?

pescuma:
Is there any problem if it always override? I'd find it strange if one build depends on the changes made by the previous one, but maybe I'm missing something.

mosabua:
I am not sure. But we are changing current behavior of the plugin so it might be safer as an option. No idea what all our users are up to..

jpricketMSFT:
Hi @pescuma, I agree with @mosabua that this should probably be a global option. At least then users can change the behavior back if there is some kind of problem. The default should probably be for it to overwrite as you have suggested.
Thanks,
Jason

pescuma:
I've agreed with @mosabua and I intended to implement this as a build option, but it kind of got lost. I'll create that option.

pescuma:
@jpricketMSFT @mosabua Done

mosabua:
I like it. Imho we should merge.

PR #153, opened 2017-03-02
Return version history without cloacked paths when listing for build

pescuma:


jpricketMSFT:
Hi @pescuma, can you explain why you would want this done and why you think everyone would want this change?
Thanks,
Jason

pescuma:
Hi @jpricketMSFT . I think that it makes sense that, if you don't want to see the contents of one folder, you also don't want to see changes in that folder, but maybe I'm influenced by how we use TFS here. Here, we have some projects inside $/ (let's say $/A , $/B and $/C) and, sometimes, to compile one you also need to compile another (let's say $/B depends on $/A). So, because we can't checkout 2 paths, I cloack the paths that are not interesting for a build (in our example, I'd download $/ and cloack $/C). So, when checking which revisions were compiled, I'm not interested in seeing revisions for $/C.

jpricketMSFT:
Hi @pescuma, Thanks for the explanation. That makes sense. I will review the code this week and provide any feedback I have. 
Thanks,
Jason

jpricketMSFT:
@pescuma - The changes look good to me. Can you fix the conflicts and push a new commit?
Thanks for this contribution!

pescuma:
@jpricketMSFT Done

PR #154, opened 2017-03-10
Allow VSTS JenkinsQueueJob to pass the branch name for multibranch pipeline job

yacaovsnc:
In v5.3.3, we are assuming the jobname for a Multibranch Pipeline job is passed from VSTS / TFS JenkinsQueueJob task in the format of _${MultibranchPipelineJob}/${Branch}_.  This won't work for Multibranch Pipeline jobs in folders as we can't pinpoint which `/` splits the job and the branch.  

In JenkinsQueueJob task 1.115.0, we are adding a separate field to specify the branch name.  This will resolve all the ambiguities.   

This fix requires JenkinsQueueJob 1.115.0.

### DIFF
--- a/tfs/src/main/java/hudson/plugins/tfs/TeamBuildEndpoint.java
+++ b/tfs/src/main/java/hudson/plugins/tfs/TeamBuildEndpoint.java
@@ -58,6 +58,7 @@ public class TeamBuildEndpoint implements UnprotectedRootAction {
     public static final String URL_NAME = "team-build";
     public static final String PARAMETER = "parameter";
     public static final String BUILD_SOURCE_BRANCH = "Build.SourceBranch";
+    public static final String QUEUEJOBTASK_MULTIBRANCH_JOB_BRANCH = "QueueJobTask.MultibranchPipelineBranch";
     static final String URL_PREFIX = "/" + URL_NAME + "/";
 
     static {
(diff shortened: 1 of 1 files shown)


tfs/src/main/java/hudson/plugins/tfs/TeamBuildEndpoint.java
```diff
@@ -196,22 +197,25 @@ void dispatch(final StaplerRequest req, final StaplerResponse rsp, final TimeDur
...
-            final TeamBuildPayload payload = EndpointHelper.MAPPER.convertValue(formData, TeamBuildPayload.class);
+        String sourceBranch = payload.BuildVariables.get(QUEUEJOBTASK_MULTIBRANCH_JOB_BRANCH);
 
-            sourceBranch = payload.BuildVariables.get(BUILD_SOURCE_BRANCH);
+        if (sourceBranch == null || sourceBranch.trim().isEmpty()) {
+            if (jobName.indexOf('/') > 0) {
+                sourceBranch = jobName.substring(jobName.indexOf('/') + 1);
+            } else {
+                sourceBranch = payload.BuildVariables.get(BUILD_SOURCE_BRANCH);
+            }
```
DavidStaheli: Not important, but consider saving the result of `jobName.indexOf('/')` instead of calling it twice.

PR #157, opened 2017-03-31
ignoring eclipse files also in the modules

mosabua:
Eclipse/M2e maintains files in each module folder and since we have a multi-module project now we need to expand the exclusion to any folder...


mosabua:
Unless there are any objections from @DavidStaheli , @olivierdagenais or @yacaovsnc I am going to merge this soon.

PR #159, opened 2017-04-22
Added the opton to toggle show workspaces in log

suratdas:
A new UI checkbox under Advanced settings gives option to show or not show the workspaces in the build log file. Sometimes, we send the build log failure by Jenkins Ext-Email plugin where we don't want to clutter log with all workspaces. In some other cases, we want to see workspaces in the log. With this configuration option, we can control each Jenkins build the way we want it.
image

### DIFF
--- a/tfs/src/main/java/hudson/plugins/tfs/TeamFoundationServerScm.java
+++ b/tfs/src/main/java/hudson/plugins/tfs/TeamFoundationServerScm.java
@@ -67,7 +67,7 @@ import static hudson.Util.fixEmpty;
 
 /**
  * SCM for Microsoft Team Foundation Server.
- * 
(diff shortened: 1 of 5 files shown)


mosabua:
I think the config should be embedded in the class TeamPluginGlobalConfig instead of in the TeamFoundationServerScm class. Wdyt @olivierdagenais ?

suratdas:
I thought TeamPluginGlobalConfig is for the options we see in the <jenkins_url>/configure screen. I had the idea of having this as configurable per build. Some builds may choose to display workspaces in log and others may choose not to. This is why I did not keep this logic in TeamPluginGlobalConfig. Let me know what you think.

PR #161, opened 2017-05-13
JENKINS-43809 Add basic proxy functionality to release triggers

JacobHenner:
This adds basic proxy functionality to release triggers. I haven't had the opportunity to test this across a range of environments, but it does work for my situation. Furthermore, there are probably a bunch of other components that also need proxy functionality added.

### DIFF
--- a/tfs/src/main/java/hudson/plugins/tfs/rm/ReleaseManagementHttpClient.java
+++ b/tfs/src/main/java/hudson/plugins/tfs/rm/ReleaseManagementHttpClient.java
@@ -1,6 +1,7 @@
 package hudson.plugins.tfs.rm;
 
 import com.google.gson.Gson;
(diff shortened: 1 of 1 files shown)


jpricketMSFT:
Hey @JacobHenner,
Your fix looks correct. In the RM code, we are incorrectly creating an HttpClient without going thru our "ModernHTTPClientFactory" which it looks like does the same thing that you are doing here with the Jenkins proxy configuration. 
I think the correct fix would be to use the factory here as well. I am not an expert on the code yet, however. If you can modify your fix to the user the factory, I will gladly merge it in. Otherwise, I will try to fix the problem as soon as I can. When I have a fix, I will merge in your fix first so you get credit for finding this bug. 
Thanks for this contribution!

PR #162, opened 2017-05-16
[JENKINS-40807] Fix Findbugs errors

varyvol:
[JENKINS-40807](https://issues.jenkins-ci.org/browse/JENKINS-40807)

After [JENKINS-40806](https://issues.jenkins-ci.org/browse/JENKINS-40806), Findbugs is run as part of the build. However, such builds does not fail if any error arise.

This change revert that behaviour and fixes/suppresses already present errors.

@reviewbybees @yacaovsnc

### DIFF
--- a/tfs/src/main/java/hudson/plugins/tfs/ChangeSetReader.java
+++ b/tfs/src/main/java/hudson/plugins/tfs/ChangeSetReader.java
@@ -7,6 +7,7 @@ import java.io.Reader;
 import java.util.ArrayList;
 import java.util.List;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import org.apache.commons.digester.Digester;
 import org.apache.commons.io.IOUtils;
 import org.xml.sax.SAXException;
@@ -25,6 +26,7 @@ import hudson.util.Digester2;
 public class ChangeSetReader extends ChangeLogParser {
 
     @Override
+    @SuppressFBWarnings(value = "DM_DEFAULT_ENCODING", justification = "Better mot modify charset in case it might raise errors")
     public ChangeLogSet parse(AbstractBuild build, File changelogFile) throws IOException, SAXException {
         FileReader reader = new FileReader(changelogFile);
         try {
--- a/tfs/src/main/java/hudson/plugins/tfs/TeamCollectResultsPostBuildAction.java
+++ b/tfs/src/main/java/hudson/plugins/tfs/TeamCollectResultsPostBuildAction.java
@@ -1,5 +1,6 @@
 package hudson.plugins.tfs;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.Extension;
 import hudson.FilePath;
 import hudson.Launcher;
@@ -51,6 +52,7 @@ public class TeamCollectResultsPostBuildAction extends Recorder implements Simpl
         this.requestedResults = requestedResults;
     }
 
+    @SuppressFBWarnings(value = "RV_RETURN_VALUE_IGNORED_BAD_PRACTICE", justification = "No matter the result of mkdirs")
     @Override
     public void perform(
             @Nonnull final Run<?, ?> run,
@@ -76,7 +78,6 @@ public class TeamCollectResultsPostBuildAction extends Recorder implements Simpl
             final String folderName = teamResultType.getFolderName();
             logger.print(" " + teamResultType.getDisplayName());
             final File resultFolder = new File(resultsRoot, folderName);
-            //noinspection ResultOfMethodCallIgnored
             resultFolder.mkdirs();
             final String includes = requestedResult.getIncludes();
             final FilePath resultPath = new FilePath(resultFolder);
--- a/tfs/src/main/java/hudson/plugins/tfs/TeamFoundationServerScm.java
+++ b/tfs/src/main/java/hudson/plugins/tfs/TeamFoundationServerScm.java
@@ -391,7 +391,7 @@ public class TeamFoundationServerScm extends SCM {
     @Override
     public boolean processWorkspaceBeforeDeletion(AbstractProject<?, ?> project, FilePath workspace, Node node) throws IOException, InterruptedException {
         Run<?,?> lastRun = project.getLastBuild();
-        if ((lastRun == null) || !(lastRun instanceof AbstractBuild<?, ?>)) {
+        if (lastRun == null) {
             return true;
         }
         
@@ -515,7 +515,6 @@ public class TeamFoundationServerScm extends SCM {
         public static final Pattern WORKSPACE_NAME_REGEX = Pattern.compile("[^\"/:<>\\|\\*\\?]+[^\\s\\.\"/:<>\\|\\*\\?]$", Pattern.CASE_INSENSITIVE);
         public static final Pattern PROJECT_PATH_REGEX = Pattern.compile("^\\$\\/.*", Pattern.CASE_INSENSITIVE);
         public static final Pattern CLOAKED_PATHS_REGEX = Pattern.compile("\\s*\\$[^\\n;]+(\\s*[\\n]\\s*\\$[^\\n;]+){0,}\\s*", Pattern.CASE_INSENSITIVE);
-        private transient String tfExecutable;
         
         public DescriptorImpl() {
             super(TeamFoundationServerScm.class, TeamFoundationServerRepositoryBrowser.class);
--- a/tfs/src/main/java/hudson/plugins/tfs/TeamPullRequestMergedDetailsAction.java
+++ b/tfs/src/main/java/hudson/plugins/tfs/TeamPullRequestMergedDetailsAction.java
@@ -45,7 +45,6 @@ public class TeamPullRequestMergedDetailsAction implements Action, Serializable 
         if (action != null && action.hasWorkItems()) {
             Collections.addAll(destination, action.getWorkItems());
             final GitPullRequestEx gitPullRequest = action.gitPullRequest;
-            final GitRepository repository = gitPullRequest.getRepository();
             final URI collectionUri = URI.create(action.collectionUri);
             return collectionUri;
         }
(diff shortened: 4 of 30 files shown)


varyvol:
@yacaovsnc did you have the opportunity to review these changes?

yacaovsnc:
+@DavidStaheli

jpricketMSFT:
Hi @varyvol, I am taking over from yacaovsnc. All of the changes seem reasonable. Unfortunately, I pushed some of my own changes before see this. So, I probably introduced some additional errors. I plan on merging your changes and then fixes my errors so we can get down to zero again. 
Thanks for this contribution!

varyvol:
@jpricketMSFT Thank you!

tfs/src/main/java/hudson/plugins/tfs/ChangeSetReader.java
```diff
@@ -25,6 +26,7 @@
 public class ChangeSetReader extends ChangeLogParser {
 
     @Override
+    @SuppressFBWarnings(value = "DM_DEFAULT_ENCODING", justification = "Better mot modify charset in case it might raise errors")
```
oleg-nenashev: Would be better to explicitly use the System default then

tfs/src/main/java/hudson/plugins/tfs/TeamCollectResultsPostBuildAction.java
```diff
@@ -51,6 +52,7 @@ public void setRequestedResults(final List<TeamRequestedResult> requestedResults
         this.requestedResults = requestedResults;
     }
 
+    @SuppressFBWarnings(value = "RV_RETURN_VALUE_IGNORED_BAD_PRACTICE", justification = "No matter the result of mkdirs")
```
oleg-nenashev: Why? This is a correct issue. Ideally a Java 7 `Files` API should be used instead

tfs/src/main/java/hudson/plugins/tfs/TeamFoundationServerScm.java
```diff
@@ -391,7 +391,7 @@ public boolean pollChanges(AbstractProject hudsonProject, Launcher launcher, Fil
     @Override
     public boolean processWorkspaceBeforeDeletion(AbstractProject<?, ?> project, FilePath workspace, Node node) throws IOException, InterruptedException {
         Run<?,?> lastRun = project.getLastBuild();
```
oleg-nenashev: You can explicitly use `AbstractBuild` here, there is no need to cast the type later

tfs/src/main/java/hudson/plugins/tfs/TeamPullRequestMergedDetailsAction.java
```diff
@@ -24,7 +24,7 @@
     private static final long serialVersionUID = 1L;
     private static final String URL_NAME = "team-pullRequestMergedDetails";
 
-    public GitPullRequestEx gitPullRequest;
+    public transient GitPullRequestEx gitPullRequest;
```
oleg-nenashev: 🐛 It is going to break the event persistence on the disk, because the field won't be stored anymore.
And this action is persisted from what I see

tfs/src/main/java/hudson/plugins/tfs/model/GitCodePushedEventArgs.java
```diff
@@ -2,10 +2,11 @@
 
 import org.eclipse.jgit.transport.URIish;
 
+import java.io.Serializable;
 import java.net.URI;
 import java.net.URISyntaxException;
 
-public class GitCodePushedEventArgs {
+public class GitCodePushedEventArgs implements Serializable {
```
oleg-nenashev: It breaks the compatibility, because the class is not final.
If you make it serializable, also define the `serialVersionID`

tfs/src/main/java/hudson/plugins/tfs/model/ManualCredentialsConfigurer.java
```diff
@@ -3,10 +3,12 @@
 import com.cloudbees.plugins.credentials.CredentialsScope;
 import com.cloudbees.plugins.credentials.common.StandardUsernamePasswordCredentials;
 import com.cloudbees.plugins.credentials.impl.UsernamePasswordCredentialsImpl;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.Extension;
 import hudson.util.Secret;
 import org.kohsuke.stapler.DataBoundConstructor;
 
+@SuppressFBWarnings(value = "SE_TRANSIENT_FIELD_NOT_RESTORED", justification = "Maintain compatibility")
```
oleg-nenashev: There should have been readResolve() then

tfs/src/main/java/hudson/plugins/tfs/util/BuildWorkspaceConfigurationRetriever.java
```diff
@@ -45,17 +45,23 @@ public BuildWorkspaceConfiguration getLatestForNode(Node needleNode, Run<?,?> la
     
     public static class BuildWorkspaceConfiguration extends WorkspaceConfiguration {
         private static final long serialVersionUID = 1L;
-        private final AbstractBuild<?, ?> build;
+        private final transient AbstractBuild<?, ?> build;
```
oleg-nenashev: I am not sure what is the use-case for this class, but it may also break persistency on the disk 🐜

tfs/src/main/java/hudson/plugins/tfs/TeamPullRequestMergedDetailsAction.java
varyvol: GitPullRequestEx is not serializable so in case someone was serializing it, it would have already broken.

tfs/src/main/java/hudson/plugins/tfs/model/GitCodePushedEventArgs.java
varyvol: Why does adding an interface break the compatibility? My understanding is that you can add them, being the removal the problematic thing.

https://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html

tfs/src/main/java/hudson/plugins/tfs/model/ManualCredentialsConfigurer.java
varyvol: Resolving to what? It's not something that can be predicted I think.

Also there are tests that are specifically checking the serialization generates an empty CredentialsConfigurer.

tfs/src/main/java/hudson/plugins/tfs/util/BuildWorkspaceConfigurationRetriever.java
varyvol: AbstractBuild is not serializable so this could have never been serialized.

varyvol:
@oleg-nenashev I have answered some of your comments. I will create another PR with some fixes based on others.

tfs/src/main/java/hudson/plugins/tfs/TeamPullRequestMergedDetailsAction.java
oleg-nenashev: "Serializable" is not required for persisting the data on the disk. My concern is that my making the class formally serializable you may break the data persistency

tfs/src/main/java/hudson/plugins/tfs/model/GitCodePushedEventArgs.java
oleg-nenashev: Serializable is a special case. If you add this interface, all child classes must become serializable as well. And there is no guarantee they really can support it.

Fromt the **binary compatibility** standpoint it is compatible, of course

tfs/src/main/java/hudson/plugins/tfs/util/BuildWorkspaceConfigurationRetriever.java
oleg-nenashev: Serialization is not required for persistency via XStream. That engine is different.
Persisting builds likely ends up with a relative link in XML, e.g. "../..". By making the field transient you are just removing it from XML at all

tfs/src/main/java/hudson/plugins/tfs/util/BuildWorkspaceConfigurationRetriever.java
```diff
@@ -45,17 +45,23 @@ public BuildWorkspaceConfiguration getLatestForNode(Node needleNode, Run<?,?> la
...
         public BuildWorkspaceConfiguration(WorkspaceConfiguration configuration, AbstractBuild<?, ?> build) {
             super(configuration);
             this.build = build;
         }
+
         public void save() throws IOException {
             if (!workspaceExists()) {
                 build.getAction(WorkspaceConfiguration.class).setWorkspaceWasRemoved();
             }
             build.save();
```
oleg-nenashev: risk of NPE after the change?

tfs/src/main/java/hudson/plugins/tfs/model/GitCodePushedEventArgs.java
jglick: Well you have to consider whether this is a plausible situation where someone might have subclassed it to begin with.

tfs/src/main/java/hudson/plugins/tfs/util/BuildWorkspaceConfigurationRetriever.java
jglick: Actually due to lazy loading issues there are cases where including a non`transient` field of type `Run` will cause serious errors, so it is mandatory to make this `transient`.

oleg-nenashev: I agree the persisted Build instance is a bad thing in any case. But the one who solves that has to ensure the plugin's logic does not suffer. Rejuice methods or whatever. It is not enough to just make the field transient

PR #163, opened 2017-06-23
Adding basic telemetry to the plugin

jpricketMSFT:
This telemetry will be sent to Application Insights for tracking what features users are engaged with.

### DIFF
--- a/tfs/src/main/java/hudson/plugins/tfs/telemetry/TelemetryContextInitializer.java
+++ b/tfs/src/main/java/hudson/plugins/tfs/telemetry/TelemetryContextInitializer.java
@@ -188,6 +188,6 @@ public class TelemetryContextInitializer implements ContextInitializer {
     }
 
     private String getPluginVersion() {
-        return getClass().getPackage().getImplementationVersion();
+        return Jenkins.getInstance().getPluginManager().getPlugin("tfs").getVersion();
     }
 }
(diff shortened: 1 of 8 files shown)


tfs/src/main/java/hudson/plugins/tfs/telemetry/TelemetryContextInitializer.java
```diff
@@ -0,0 +1,193 @@
...
+public class TelemetryContextInitializer implements ContextInitializer {
+    private static final Logger logger = LoggerFactory.getLogger(TelemetryContextInitializer.class);
+
+    private static final String PRODUCT_NAME= "TFS-Jenkins";
+    private static final String SYS_PROP_OS_NAME = "os.name";
+    private static final String SYS_PROP_OS_VERSION = "os.version";
+    private static final String SYS_PROP_USER_NAME = "user.name";
+    private static final String SYS_PROP_JAVA_RUNTIME = "java.runtime.name";
+    private static final String SYS_PROP_JAVA_VERSION = "java.version";
+    private static final String USER_AGENT_FORMAT = "{0}/{1} {2}/{3} {4}/{5} {6}/{7} ({9})";
```
jeffyoung: What do have against 8?

jpricketMSFT: Good catch! 9 should be 8.

jpricketMSFT:
I agree. There are probably a lot more places we need telemetry. I just wasn't sure where those other places were. This initial telemetry is mostly a guess on the places I think are most interesting. I could be way off. As for error conditions, I just don't know which "catches" are exceptional yet and which aren't :(

PR #165, opened 2017-06-27
[JENKINS-40807] Fix and improve removal of Findbugs errors

varyvol:
[JENKINS-40807](https://issues.jenkins-ci.org/browse/JENKINS-40807)

After [PR #162](https://github.com/jenkinsci/tfs-plugin/pull/162) was merged, some comments were made in the PR. This fixes/improves some of the changes there with such proposals.

@jpricketMSFT @oleg-nenashev

### DIFF
--- a/tfs/src/main/java/hudson/plugins/tfs/TeamCollectResultsPostBuildAction.java
+++ b/tfs/src/main/java/hudson/plugins/tfs/TeamCollectResultsPostBuildAction.java
@@ -1,6 +1,5 @@
 package hudson.plugins.tfs;
 
-import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.Extension;
 import hudson.FilePath;
 import hudson.Launcher;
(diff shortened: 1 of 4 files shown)


tfs/src/main/java/hudson/plugins/tfs/TeamCollectResultsPostBuildAction.java
```diff
@@ -82,7 +80,7 @@ public void perform(
             final String folderName = teamResultType.getFolderName();
             logger.print(" " + teamResultType.getDisplayName());
             final File resultFolder = new File(resultsRoot, folderName);
-            resultFolder.mkdirs();
+            Files.createDirectory(resultFolder.toPath());
```
jpricketMSFT: To be consistent with the previous behavior, I think you should use Files.createDirectories (since you are replacing mkdirs). This will create any parent directories that don't exist.

varyvol: Absolutely! I thought this was createDirectory's behaviour.

PR #167, opened 2017-08-10
Adding JenkinsEventNotifier to send events back to TFS/VSTS

jpricketMSFT:
Adding "team-events-connect" endpoint which allows TFS/VSTS to "connect" to the Jenkins server and start receiving events.

### DIFF
--- a/tfs/src/main/java/hudson/plugins/tfs/JenkinsEventNotifier.java
+++ b/tfs/src/main/java/hudson/plugins/tfs/JenkinsEventNotifier.java
@@ -40,19 +40,16 @@ public class JenkinsEventNotifier {
             try {
                 // Check to see if there are any collections "connected" to this Jenkins server
                 final ConnectionParameters connectionParameters = c.getConnectionParameters();
-                if (connectionParameters == null || !connectionParameters.isSendJobCompletionEvents()) {
-                    // This server isn't accepting events
-                    continue;
-                }
                 final TeamRestClient client = new TeamRestClient(URI.create(c.getCollectionUrl()));
                 payload.put("server", connectionParameters.getConnectionKey());
                 final String jsonPayload = payload.toString();
                 final JobCompletionEventArgs args = new JobCompletionEventArgs(
-                        connectionParameters.getConnectionKey(), jsonPayload,
+                        connectionParameters.getConnectionKey(),
+                        jsonPayload,
                         getPayloadSignature(connectionParameters, jsonPayload));
                 client.sendJobCompletionEvent(args);
             } catch (final Exception e) {
-                log.info("ERROR: sendJobCompletionEvent: (collection=" + c.getCollectionUrl() + ") " + e.getMessage());
+                log.warning("ERROR: sendJobCompletionEvent: (collection=" + c.getCollectionUrl() + ") " + e.getMessage());
             }
         }
     }
@@ -85,20 +82,22 @@ public class JenkinsEventNotifier {
             }
             return result.toString();
         } catch (final IOException e) {
-            log.info("ERROR: getApiJson: (url=" + url + ") " + e.getMessage());
+            log.warning("ERROR: getApiJson: (url=" + url + ") " + e.getMessage());
             throw new RuntimeException(e);
         }
     }
 
     private static String urlCombine(final String url, final String... parts) {
         final StringBuilder sb = new StringBuilder();
-        sb.append(url);
-        for(final String s : parts) {
-            if (StringUtils.isNotBlank(s)) {
-                if (sb.charAt(sb.length() - 1) != '/') {
-                    sb.append("/");
+        if (url != null) {
+            sb.append(url);
+            for (final String s : parts) {
+                if (StringUtils.isNotBlank(s)) {
+                    if (sb.length() > 0 && sb.charAt(sb.length() - 1) != '/') {
+                        sb.append("/");
+                    }
+                    sb.append(s);
                 }
-                sb.append(s);
             }
         }
         return sb.toString();
--- a/tfs/src/main/java/hudson/plugins/tfs/JenkinsEventNotifier.java
+++ b/tfs/src/main/java/hudson/plugins/tfs/JenkinsEventNotifier.java
@@ -1,0 +1,126 @@
+package hudson.plugins.tfs;
+
+import hudson.plugins.tfs.model.ConnectionParameters;
+import hudson.plugins.tfs.model.JobCompletionEventArgs;
+import hudson.plugins.tfs.util.TeamRestClient;
+import jenkins.model.Jenkins;
+import net.sf.json.JSONObject;
+import org.apache.commons.lang.StringUtils;
+import org.apache.http.HttpResponse;
+import org.apache.http.client.HttpClient;
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.impl.client.HttpClientBuilder;
+
+import javax.crypto.Mac;
+import javax.crypto.spec.SecretKeySpec;
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.net.URI;
+import java.security.InvalidKeyException;
+import java.security.NoSuchAlgorithmException;
+import java.util.Formatter;
+import java.util.List;
+import java.util.logging.Logger;
+
+/**
+ * This class sends Jenkins events to all "connected" team collections.
+ * Current list of events:
+ * - Job Completion event
+ */
+public class JenkinsEventNotifier {
+    protected final static Logger log = Logger.getLogger(JenkinsEventNotifier.class.getName());
+
+    /**
+     * Send the Job Completion event to connected TFS/VSTS servers
+     */
+    public static void sendJobCompletionEvent(final JSONObject payload) {
+        final List<TeamCollectionConfiguration> connectedCollections = TeamCollectionConfiguration.getConnectedCollections();
+        for (final TeamCollectionConfiguration c : connectedCollections) {
+            try {
+                // Check to see if there are any collections "connected" to this Jenkins server
+                final ConnectionParameters connectionParameters = c.getConnectionParameters();
+                if (connectionParameters == null || !connectionParameters.isSendJobCompletionEvents()) {
+                    // This server isn't accepting events
+                    continue;
+                }
+                final TeamRestClient client = new TeamRestClient(URI.create(c.getCollectionUrl()));
+                payload.put("server", connectionParameters.getConnectionKey());
+                final String jsonPayload = payload.toString();
+                final JobCompletionEventArgs args = new JobCompletionEventArgs(
+                        connectionParameters.getConnectionKey(), jsonPayload,
+                        getPayloadSignature(connectionParameters, jsonPayload));
+                client.sendJobCompletionEvent(args);
+            } catch (final Exception e) {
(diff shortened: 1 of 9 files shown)


tfs/src/main/java/hudson/plugins/tfs/JenkinsEventNotifier.java
```diff
@@ -0,0 +1,126 @@
...
+import java.util.List;
+import java.util.logging.Logger;
+
+/**
+ * This class sends Jenkins events to all "connected" team collections.
+ * Current list of events:
+ * - Job Completion event
+ */
+public class JenkinsEventNotifier {
+    protected final static Logger log = Logger.getLogger(JenkinsEventNotifier.class.getName());
```
lkillgore: (I don't know if we care): I think Checkstyle will flag this line for the order of the keywords:

https://stackoverflow.com/questions/10299067/modifier-keyword-order-in-java

tfs/src/main/java/hudson/plugins/tfs/JenkinsEventNotifier.java
```diff
@@ -0,0 +1,126 @@
...
+                final ConnectionParameters connectionParameters = c.getConnectionParameters();
+                if (connectionParameters == null || !connectionParameters.isSendJobCompletionEvents()) {
+                    // This server isn't accepting events
+                    continue;
+                }
+                final TeamRestClient client = new TeamRestClient(URI.create(c.getCollectionUrl()));
+                payload.put("server", connectionParameters.getConnectionKey());
+                final String jsonPayload = payload.toString();
+                final JobCompletionEventArgs args = new JobCompletionEventArgs(
+                        connectionParameters.getConnectionKey(), jsonPayload,
```
AlexRukhlin: Split this line ?

tfs/src/main/java/hudson/plugins/tfs/JenkinsEventNotifier.java
```diff
@@ -0,0 +1,126 @@
...
+            throw new RuntimeException(e);
+        }
+    }
+
+    private static String urlCombine(final String url, final String... parts) {
+        final StringBuilder sb = new StringBuilder();
+        sb.append(url);
+        for(final String s : parts) {
+            if (StringUtils.isNotBlank(s)) {
+                if (sb.charAt(sb.length() - 1) != '/') {
```
lkillgore: Protect against an empty url?

tfs/src/main/java/hudson/plugins/tfs/JenkinsEventNotifier.java
jpricketMSFT: Probably right. I will run check style and fix those issues.

tfs/src/main/java/hudson/plugins/tfs/JenkinsEventNotifier.java
jpricketMSFT: I assume you want me to put each parameter on a separate line? That makes sense.

tfs/src/main/java/hudson/plugins/tfs/JenkinsEventNotifier.java
jpricketMSFT: good catch.

tfs/src/main/java/hudson/plugins/tfs/listeners/JenkinsRunListener.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.logging.Logger;
+
+/**
+ * This class listens to the events of every Jenkins run instance.
+ * Completed runs fire an event back to the JenkinsEventNotifier.
+ */
+@Extension
+public class JenkinsRunListener extends RunListener<Run> {
+
+    protected static Logger log = Logger.getLogger(JenkinsRunListener.class.getName());
```
lkillgore: +final

tfs/src/main/java/hudson/plugins/tfs/JenkinsEventNotifier.java
AlexRukhlin: Shouldn't we check that sb.Length() > 0 ?

tfs/src/main/java/hudson/plugins/tfs/listeners/JenkinsRunListener.java
```diff
@@ -0,0 +1,59 @@
...
+        final JSONObject json = JSONObject.fromObject(payload);
+        json.put("name", run.getParent().getDisplayName());
+        json.put("startedBy", getStartedBy(run));
+
+        JenkinsEventNotifier.sendJobCompletionEvent(json);
+    }
+
+    private String getStartedBy(final Run run) {
+        final Cause.UserIdCause cause = (Cause.UserIdCause) run.getCause(Cause.UserIdCause.class);
+        final String startedBy;
```
lkillgore: I think this would be simpler without the 'final' and the 'else' clause.  Just initialize to 'null'.

tfs/src/main/java/hudson/plugins/tfs/JenkinsEventNotifier.java
```diff
@@ -0,0 +1,126 @@
...
+    /**
+     * Send the Job Completion event to connected TFS/VSTS servers
+     */
+    public static void sendJobCompletionEvent(final JSONObject payload) {
+        final List<TeamCollectionConfiguration> connectedCollections = TeamCollectionConfiguration.getConnectedCollections();
+        for (final TeamCollectionConfiguration c : connectedCollections) {
+            try {
+                // Check to see if there are any collections "connected" to this Jenkins server
+                final ConnectionParameters connectionParameters = c.getConnectionParameters();
+                if (connectionParameters == null || !connectionParameters.isSendJobCompletionEvents()) {
```
AlexRukhlin: It seems to me that the method TeamCollectionConfiguration.getConnectedCollections() already does this check.

jpricketMSFT: good point. I will remove the null check and perhaps annotate the method with NotNull

tfs/src/main/java/hudson/plugins/tfs/JenkinsEventNotifier.java
jpricketMSFT: absolutely, good catch.

tfs/src/main/java/hudson/plugins/tfs/listeners/JenkinsRunListener.java
jpricketMSFT: yep

tfs/src/main/java/hudson/plugins/tfs/TeamEventsEndpoint.java
```diff
@@ -158,8 +160,8 @@ static JSONObject innerDispatch(final String body, final String eventName, final
         }
         final AbstractHookEvent.Factory factory = factoriesByName.get(eventName);
         final Event serviceHookEvent = deserializeEvent(body);
-        final String message = serviceHookEvent.getMessage().getText();
-        final String detailedMessage = serviceHookEvent.getDetailedMessage().getText();
+        final String message = serviceHookEvent.getMessage() != null ? serviceHookEvent.getMessage().getText() : "";
+        final String detailedMessage = serviceHookEvent.getDetailedMessage() != null ? serviceHookEvent.getDetailedMessage().getText() : "";
```
AlexRukhlin: Why do we need these messages?

tfs/src/main/java/hudson/plugins/tfs/listeners/JenkinsRunListener.java
jpricketMSFT: It would be less lines, but I like the final. I will see how it looks.

tfs/src/main/java/hudson/plugins/tfs/util/TeamRestClient.java
```diff
@@ -287,4 +305,27 @@ public TeamGitStatus addPullRequestIterationStatus(final PullRequestMergeCommitC
 
         return request(TeamGitStatus.class, HttpMethod.POST, requestUri, status);
     }
+
+
+    public void sendJobCompletionEvent(final JobCompletionEventArgs args) throws IOException {
+        final QueryString qs = new QueryString(
+                API_VERSION, "3.0-preview.1",
```
AlexRukhlin: 4.0 ?

tfs/src/main/java/hudson/plugins/tfs/TeamEventsEndpoint.java
AlexRukhlin: It looks like we do not use these messages.

tfs/src/main/java/hudson/plugins/tfs/util/TeamRestClient.java
AlexRukhlin: Shouldn't we use a released API version?

tfs/src/main/java/hudson/plugins/tfs/JenkinsEventNotifier.java
```diff
@@ -0,0 +1,126 @@
...
+                }
+                final TeamRestClient client = new TeamRestClient(URI.create(c.getCollectionUrl()));
+                payload.put("server", connectionParameters.getConnectionKey());
+                final String jsonPayload = payload.toString();
+                final JobCompletionEventArgs args = new JobCompletionEventArgs(
+                        connectionParameters.getConnectionKey(), jsonPayload,
+                        getPayloadSignature(connectionParameters, jsonPayload));
+                client.sendJobCompletionEvent(args);
+            } catch (final Exception e) {
+                log.info("ERROR: sendJobCompletionEvent: (collection=" + c.getCollectionUrl() + ") " + e.getMessage());
```
yacaovsnc: Maybe this log level should be "warn"?  Is this error expected?

tfs/src/main/java/hudson/plugins/tfs/JenkinsEventNotifier.java
jpricketMSFT: I added CheckStyle to the build but it will be in another PR since there are so many changes.

tfs/src/main/java/hudson/plugins/tfs/JenkinsEventNotifier.java
jpricketMSFT: Agreed. It is not expected. I looked for a log.error, but there wasn't one :(

tfs/src/main/java/hudson/plugins/tfs/TeamEventsEndpoint.java
jpricketMSFT: I don't see them being used either, but it would be bigger change to remove them. I only changed the code because I ran into an error here. I don't know why they were originally added.

tfs/src/main/java/hudson/plugins/tfs/listeners/JenkinsRunListener.java
jpricketMSFT: I made this change in another PR (with CheckStyle)

tfs/src/main/java/hudson/plugins/tfs/util/TeamRestClient.java
jpricketMSFT: will do. Thanks for providing the version information.

PR #174, opened 2017-08-23
Remove Codeplex instructions from README

nschonni:
https://blogs.msdn.microsoft.com/bharry/2017/03/31/shutting-down-codeplex/
> So, it’s time to say goodbye to CodePlex.  As of this post, we’ve disabled the ability to create new CodePlex projects.  In October, we’ll set CodePlex to read-only, before shutting it down completely on December 15th, 2017.

There are some tests that should also get removed, but I'm just doing this from the web UI
https://github.com/jenkinsci/tfs-plugin/search?utf8=%E2%9C%93&q=codeplex&type=

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,7 +156,7 @@ If your source code is in a TFVC repository, this section is for you.
 
 Field | Description
 ----- | -----------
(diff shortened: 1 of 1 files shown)


nschonni:
CodePlex has been converted to Archive mode now

PR #177, opened 2017-09-26
Use default values for parameters when value not supplied

kelliejos:
Updated the build command to check for any parameters that have not been defined and uses their default value.

### DIFF
--- a/tfs/src/main/java/hudson/plugins/tfs/model/BuildCommand.java
+++ b/tfs/src/main/java/hudson/plugins/tfs/model/BuildCommand.java
@@ -145,10 +145,12 @@ public class BuildCommand extends AbstractCommand {
         if (pp != null && requestPayload.containsKey(TeamBuildEndpoint.PARAMETER)) {
             final List<ParameterValue> values = new ArrayList<ParameterValue>();
             final JSONArray a = requestPayload.getJSONArray(TeamBuildEndpoint.PARAMETER);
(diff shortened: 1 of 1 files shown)


tfs/src/main/java/hudson/plugins/tfs/model/BuildCommand.java
```diff
@@ -179,6 +181,13 @@ else if (name.equals(PULL_REQUEST_ID) && pullRequestId != null && d instanceof S
                 }
             }
 
+            //Pick up default build parameters that have not been overridden
+            for(ParameterDefinition paramDef : pp.getParameterDefinitions()) {
```
jpricketMSFT: add a final inside the for

tfs/src/main/java/hudson/plugins/tfs/model/BuildCommand.java
```diff
@@ -145,10 +145,12 @@ else if ("git.pullrequest.merged".equals(eventType)) {
         if (pp != null && requestPayload.containsKey(TeamBuildEndpoint.PARAMETER)) {
```
jpricketMSFT: Can you now remove the "//noinspection" line above? I am not sure, but it seems like you should be able to since you are using pp later on.

kelliejos: Done.

PR #178, opened 2017-09-27
Add support to expand default job params using env vars like ${GitRepoUrl}

smile21prc:
1. Add support to expand default job params using env vars like ${GitRepoUrl}
2. Property handle a couple of corner cases.


smile21prc:
Hi, @mmitche @jpricketMSFT, still can't add you as reviewers due to inadequate access. These are additional changes to support to expand default job params using env vars like ${GitRepoUrl} to support triggering of auto-generated jobs.

tfs/src/main/java/hudson/plugins/tfs/TeamPushTrigger.java
```diff
@@ -76,6 +76,14 @@ public void setJobContext(final String jobContext) {
         this.jobContext = jobContext;
     }
 
+    public Job<?, ?> getJob() {
+        return this.job;
+    }
+
+    public void setJob(final Job<?, ?> job) {
```
jpricketMSFT: It seems wrong to have a setter for Job in this class. I don't see it being used any way.

tfs/src/main/java/hudson/plugins/tfs/model/AbstractHookEvent.java
```diff
@@ -144,7 +143,7 @@ static JSONObject fromResponseContributors(final List<GitStatus.ResponseContribu
                     final CauseAction causeAction = new CauseAction(cause);
                     final Action[] actionArray = ActionHelper.create(actionsWithSafeParams, causeAction);
                     scmTriggerItem.scheduleBuild2(quietPeriod, actionArray);
-                    if (gitCodePushedEventArgs instanceof PullRequestMergeCommitCreatedEventArgs) {
+                    if (gitCodePushedEventArgs instanceof PullRequestMergeCommitCreatedEventArgs && !gitCodePushedEventArgs.collectionUri.toString().equalsIgnoreCase(TESTING_COLLECTION)) {
```
jpricketMSFT: I don't understand the need to exclude the "TESTING_COLLECTION" here

tfs/src/main/java/hudson/plugins/tfs/TeamPushTrigger.java
jpricketMSFT: I see it being used now. It still seems strange that you can change the Job that the trigger was created with.

tfs/src/main/java/hudson/plugins/tfs/model/AbstractHookEvent.java
```diff
@@ -275,15 +277,32 @@ private Boolean branchMatches(final GitCodePushedEventArgs gitCodePushedEventArg
...
 
+                this.envVars = project instanceof Job ? buildEnv((Job) project) : new EnvVars();
+                List<Action> actionsWithSafeParams = actions;
+                if (project instanceof Job) {
+                    // Add job-specific additional varaiables to current run.
+                    final ArrayList<ParameterValue> values = getDefaultParameters((Job) project);
+                    final String vstsRefspec = getVstsRefspec(gitCodePushedEventArgs);
+                    values.add(new StringParameterValue("vstsRefspec", vstsRefspec));
+                    values.add(new StringParameterValue("vstsBranchOrCommit", gitCodePushedEventArgs.commit));
+                    SafeParametersAction paraAction = new SafeParametersAction(values);
```
jpricketMSFT: final?

tfs/src/main/java/hudson/plugins/tfs/model/AbstractHookEvent.java
```diff
@@ -294,15 +313,24 @@ private Boolean branchMatches(final GitCodePushedEventArgs gitCodePushedEventArg
                     for (final RemoteConfig repository : git.getRepositories()) {
                         boolean repositoryMatches = false;
                         for (final URIish remoteURL : repository.getURIs()) {
-                            if (UriHelper.areSameGitRepo(uri, remoteURL)) {
+                            String url = this.envVars.expand(remoteURL.toString());
```
jpricketMSFT: final?

tfs/src/main/java/hudson/plugins/tfs/model/AbstractHookEvent.java
```diff
@@ -294,15 +313,24 @@ private Boolean branchMatches(final GitCodePushedEventArgs gitCodePushedEventArg
...
                         boolean repositoryMatches = false;
                         for (final URIish remoteURL : repository.getURIs()) {
-                            if (UriHelper.areSameGitRepo(uri, remoteURL)) {
+                            String url = this.envVars.expand(remoteURL.toString());
+                            URIish urlWithExpandedParams = null;
+                            try {
+                                urlWithExpandedParams = new URIish(url);
+                            } catch (URISyntaxException e) {
+                                LOGGER.warning(String.format("Remote URL '%s' fails to be re-created post build env expansion due to error: '%s'", url, e.toString()));
+                                continue;
```
jpricketMSFT: Seems like if expansion fails you can still do the comparison with the unexpanded value. Perhaps the catch clause should just log it, "urlWithExpandedParams = remoteURL" and let the remaining code execute.

tfs/src/main/java/hudson/plugins/tfs/model/AbstractHookEvent.java
```diff
@@ -275,15 +277,32 @@ private Boolean branchMatches(final GitCodePushedEventArgs gitCodePushedEventArg
                     continue;
                 }
 
+                this.envVars = project instanceof Job ? buildEnv((Job) project) : new EnvVars();
```
jpricketMSFT: seems like the check for instance of job could be pushed down into buildEnv. Not a big deal, just makes this code more readable.

tfs/src/main/java/hudson/plugins/tfs/model/AbstractHookEvent.java
```diff
@@ -371,4 +399,23 @@ private String getVstsRefspec(final GitCodePushedEventArgs gitCodePushedEventArg
             return "+refs/heads/*:refs/remotes/origin/*";
         }
     }
+
+    private EnvVars buildEnv(final Job<?, ?> job) {
+        EnvVars env = new EnvVars();
```
jpricketMSFT: final?

tfs/src/main/java/hudson/plugins/tfs/model/AbstractHookEvent.java
```diff
@@ -371,4 +399,23 @@ private String getVstsRefspec(final GitCodePushedEventArgs gitCodePushedEventArg
...
+                contributor.buildEnvironmentFor(job, env, TaskListener.NULL);
+            } catch (Exception e) {
+                LOGGER.warning("failed to build env, skipping");
+            }
+        }
+        return env;
+    }
+
+    private EnvVars addVarsToBuildEnv(final EnvVars envs, final ArrayList<ParameterValue> vars) {
+        for (ParameterValue p : vars) {
```
jpricketMSFT: final?

tfs/src/main/java/hudson/plugins/tfs/TeamPushTrigger.java
smile21prc: You are right, just need to be cautions job isn't null when TeamPushTrigger is initialized. I checked all the callers, and it's the case, so both setters and getters of jobs are unnecessary for now. Removed them.

tfs/src/main/java/hudson/plugins/tfs/model/AbstractHookEvent.java
smile21prc: Good catch. Interesting CHECKSTYLE didn't report in this case. Added final to it.

tfs/src/main/java/hudson/plugins/tfs/model/AbstractHookEvent.java
smile21prc: CheckStyle overlooked this case as well -- perhaps the logic around it make it confused whether this var was modified later or not.

tfs/src/main/java/hudson/plugins/tfs/model/AbstractHookEvent.java
smile21prc: Fixed them.

tfs/src/main/java/hudson/plugins/tfs/model/AbstractHookEvent.java
smile21prc: Fixed it.

tfs/src/main/java/hudson/plugins/tfs/model/AbstractHookEvent.java
smile21prc: Good suggestion. I considered that before, but it needed to change buildEnv to take "project" type instead "Job" type. As its internal logic is only applicable for a Job, the input type of "Job" seemed more appropriate. Luckily it's just one line of difference and doesn't look too bad with current approach. :)

tfs/src/main/java/hudson/plugins/tfs/model/AbstractHookEvent.java
smile21prc: If the try part fails, likely urlWithExpandedParams will be null and the areSameGitRepo fails, but as it's just one line of difference, sure removed "continue;" to let the remaining code execute as suggested. Changed the logging part as well.

tfs/src/main/java/hudson/plugins/tfs/model/AbstractHookEvent.java
smile21prc: The exclusion is necessary as TESTING_COLLECTION is a fake collection and its url isn't accessible. When sendPullRequestBuildStatusEvent tries to query TESTING_COLLECTION it fails due to auth issue. Excluding it as we don't need to actually reach the testing collection.

smile21prc:
Thanks @jpricketMSFT for the review. I've updated based on your comments. Would you like to take another look? Thanks.

tfs/src/main/java/hudson/plugins/tfs/model/AbstractHookEvent.java
smile21prc: Actually I changed to update queue status in a different location. Now we no longer need "TESTING_COLLECTION", and I've hence removed it.

jpricketMSFT:
@smile21prc can you rebase and push your changes again. We fixed a build break last week that caused a lot of PRs to fail validation.
Thanks, Jason

PR #179, opened 2017-10-04
Prevent HttpConnectionError when getting JSON for Jenkins object

kelliejos:
Add code to explicitly require either TLS 1.1 or 1.2. Also added a separate catch for the HttpHostConnectException to prevent it from being rethrown.

### DIFF
--- a/tfs/src/main/java/hudson/plugins/tfs/JenkinsEventNotifier.java
+++ b/tfs/src/main/java/hudson/plugins/tfs/JenkinsEventNotifier.java
@@ -9,12 +9,15 @@ import org.apache.commons.lang.StringUtils;
 import org.apache.http.HttpResponse;
 import org.apache.http.client.HttpClient;
 import org.apache.http.client.methods.HttpGet;
+import org.apache.http.conn.HttpHostConnectException;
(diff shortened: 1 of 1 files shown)


tfs/src/main/java/hudson/plugins/tfs/JenkinsEventNotifier.java
```diff
@@ -74,6 +79,17 @@ public static String getApiJson(final String url) {
             }
             final String rootUrl = jenkins.getRootUrl();
             final String fullUrl = urlCombine(rootUrl, url, "api", "json");
+            SSLContext sslContext = SSLContexts.custom()
+                    .useTLS()
+                    .build();
+
+            SSLConnectionSocketFactory f = new SSLConnectionSocketFactory(
+                    sslContext,
+                    new String[]{"TLSv1.1", "TLSv1.2"},
```
jpricketMSFT: Are there constants for TLSv1.1 and 1.2 strings?

kelliejos: There aren't. I didn't add any as these values aren't used elsewhere. That being said, it could certainly be added.

PR #181, opened 2017-10-13
Allow users to map sub-folders of cloaked paths

pvandervelde:
This pull request adds the ability to cloak a folder and then explicitly map a sub folder, so that the cloaked folder will not be included in the checkout but the sub folder will be

This capability is useful in cases (like ours) where a TFS repository contains a large amount of code and each build only needs a 'small' amount of that code. For instance if the TFS repository looks like

    $/
        product-1
            folder-1
            folder-2
        product-2
            folder-1
            folder-2
        shared-1
            folder-1
                folder-1
                folder-2
            folder-2
                folder-1
                folder-2
        shared-2
            folder-1
                folder-1
                folder-2
            folder-2
                folder-1
                folder-2

To build `product-1` the following folders are required

    $/
        product-1
            folder-1
            folder-2
        shared-1
            folder-1
                folder-1
                folder-2
        shared-2
            folder-1

To build `product-2` the following folders are required

    $/
        product-2
            folder-1
            folder-2
        shared-1
            folder-1
                folder-2
            folder-2
                folder-1
                folder-2
        shared-2
            folder-2
                folder-1
                folder-2

The build configuration for `product-1` would ideally only checkout the folders it needs, however the
`project path` has to be `$/` (i.e. the repository root). It is then possible to cloak the folders that shouldn't be included. However that means that if new folders are added the build configuration needs to be updated to cloak the new folder.

By adding the mapping ability it is possible to cloak all [top level folders](teamfoundation.blogspot.co.nz/2008/06/advanced-workspace-cloaking-in-tfs-2008.html) and explicitly map those folders that we need. That way it is unlikely that the addition of new folders will require updates to the build configuration.

Additionally this code allows mapping a folder to a different location in the workspace which enables
a reduce the path length, especially when dealing with nested branches like this:

    $/
        Branches
            1.0
                product-1
                    ...
                product-2
                    ...
            2.0
                product-1
                    ...
                product-2
                    ...
            3.0
                product-1
                    ...
                product-2
                    ...

### DIFF
--- a/tfs/src/main/java/hudson/plugins/tfs/TeamFoundationServerScm.java
+++ b/tfs/src/main/java/hudson/plugins/tfs/TeamFoundationServerScm.java
@@ -429,7 +429,7 @@ public class TeamFoundationServerScm extends SCM {
             final Project project = server.getProject(projPath);
             final int changeSet = recordWorkspaceChangesetVersion(build, listener, project, projPath, singleVersionSpec);
 
-            CheckoutAction action = new CheckoutAction(workspaceConfiguration.getWorkspaceName(), workspaceConfiguration.getProjectPath(), workspaceConfiguration.getCloakedPaths(), workspaceConfiguration.getWorkfolder(), isUseUpdate(), isUseOverwrite());
+            CheckoutAction action = new CheckoutAction(workspaceConfiguration.getWorkspaceName(), workspaceConfiguration.getProjectPath(), workspaceConfiguration.getCloakedPaths(), workspaceConfiguration.getMappedPaths(), workspaceConfiguration.getWorkfolder(), isUseUpdate(), isUseOverwrite());
             List<ChangeSet> list;
             if (StringUtils.isNotEmpty(singleVersionSpec)) {
                 list = action.checkoutBySingleVersionSpec(server, workspaceFilePath, singleVersionSpec);
@@ -502,7 +502,8 @@ public class TeamFoundationServerScm extends SCM {
                 return (server.getProject(getProjectPath(lastRun)).getDetailedHistoryWithoutCloakedPaths(
                             lastRun.getTimestamp(),
                             Calendar.getInstance(),
-                            getCloakedPaths(lastRun)
+                            getCloakedPaths(lastRun),
+                            getMappedPaths(lastRun).keySet()
                         ).size() > 0);
             } finally {
                 server.close();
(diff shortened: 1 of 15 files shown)


pvandervelde:
This is currently being tested in our test environment. I have tested with a fake TFS repository (i.e. a proper TFS server but not one of our product repositories) and it all works. We will be testing more and if it works we will start running it in our production environment in the next few weeks.

jpricketMSFT:
@pvandervelde can you fix the merge conflicts. We want to take your changes as part of the next release. Thanks, Jason

pvandervelde:
@jpricketMSFT I've rebased the branch and fixed all the merge conflicts. Looks like there's one test error on Linux but I'm pretty sure I didn't touch that code so I'm not sure why it is failing now.

pvandervelde:
@jpricketMSFT @kelliejos Rebased again on master after last release. Let me know if there's anything else I need to do for this to go in to the next release. Thanks

jpricketMSFT:
@pvandervelde, we will get this merged as soon as we can. Sorry we didn't get it in before the last release.

pvandervelde:
@jpricketMSFT Any idea when the next release will go out?

jpricketMSFT:
With the holidays starting, we don't have anything planned for the next couple of weeks. I will see if we can speed things up a bit and get out a release soon.

sitofabi:
Hi,
I've tried this pull request and really like it but I have a small issue with using it inside a pipeline. If you use the snippet generator and add only one MappedPaths it works. If you add a second entry the mappedPaths entry in the generated pipeline is gone.

Furthermore if i run a pipeline with multiple entries in mappedPaths i receive the following exception

`java.lang.ArrayIndexOutOfBoundsException: 1
	at hudson.plugins.tfs.TeamFoundationServerScm.deserializeMappedPathCollectionFromString(TeamFoundationServerScm.java:384)
	at hudson.plugins.tfs.TeamFoundationServerScm.setMappedPaths(TeamFoundationServerScm.java:248)`

As far as I can see in the TeamFoundationServerScm.java the check done at line 400 is missing in line 384. Unfortunately I'm not very familliar with java and maven so i'm not able to fix this but could you please verify this

pvandervelde:
@sitofabi Nice catch! I just pushed some changes that verify that the mapping actually contains a local path and if not we now ignore it because that's a legal thing to do. With these changes I think your use case should now work, although I am not sure how the pipeline stuff works.

Thanks for reporting the error.

pvandervelde:
@jpricketMSFT Any idea when the next release will go out. Would love to not have to run our own local version of the plugin.

pvandervelde:
@jpricketMSFT Rebased again on the latest changes in master.

pvandervelde:
Woot!

PR #184, opened 2017-11-09
Updating unit test to only run only windows OS

kelliejos:
The upgradeFromScrambledPassword() test was failing on Linux machines. Updating the test to only run on Windows OS.

### DIFF
--- a/tfs/src/test/java/hudson/plugins/tfs/TeamFoundationServerScmTest.java
+++ b/tfs/src/test/java/hudson/plugins/tfs/TeamFoundationServerScmTest.java
@@ -1,13 +1,22 @@
 package hudson.plugins.tfs;
 
-import static org.hamcrest.CoreMatchers.is;
-import static org.junit.Assert.*;
-import static org.mockito.Matchers.isA;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.times;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.verifyNoMoreInteractions;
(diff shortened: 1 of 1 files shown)


mosabua:
This might be a naive question.. but shouldn't this work on Linux too? Plenty of people use Linux nodes so from all I understand this should work on Linux.

mosabua:
Could I get an answer why it is okay to skip the test on Linux please?

jpricketMSFT:
@mosabua - One of the asserts in the test fails on linux because the scrambled password is not being returned exactly the same way every time. We might be able to remove the assert, but we didn't want to change the test significantly since we didn't write it. If you think changing/removing the test is a better solution, we will reconsider the change.

stefandrissen:
[hudson.util.secret](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/util/Secret.java) is being used to encrypt / decrypt the password - if I read the comment correctly ("This is not meant as a protection against code running in the same VM, nor against an attacker who has local file system access on Jenkins master."), the test on Linux is simply failing because it is a different VM - if you were to use a different Windows slave for the build I would expect the test would also fail.

jpricketMSFT:
@stefandrissen that makes sense. It sounds like the assert in the test is not really doing what we wanted it to. I will try to fix the test so that it can run on any node/VM

mosabua:
Thanks ... I knew something is going on that we should understand better. Thanks for chiming in @stefandrissen

mosabua:
Also btw @kelliejos ... can you please clean up branches after merges?

PR #185, opened 2017-11-09
prepare for release 5.126.0

kelliejos:


mosabua:
Why have we abandoned semantic versions and ended up wth 5.126.0 as version. Not a good approach imho.

jpricketMSFT:
We number our sprints. So we use our sprint number as the minor version. This allows us to easily know when that release was made and correlate with our development processes. I believe this still adheres to the Semantic Versioning guidelines found here: http://semver.org. The fact that the minor version does not always increment by one does not mean that this is not a semantic version number. In fact, the only guidelines that suggest what the numbers should start at involve the major and patch version numbers. There is no suggested starting value for the minor version.

PR #187, opened 2017-11-15
Fix commit list for pipeline script from SCM

kvarec:


kvarec:
Can someone add this changes to the next version?

PR #189, opened 2017-12-05
Release configuration and archiving VSTS logs improvement

zackliu:
* Support using credentials Id and dropdown list to choose Team project and Release definition.
* Auto generate credentials (if cannot using the existing one) for backwards compatibility.
* Add a release logs link in build page.

image
image


tfs/src/main/java/hudson/plugins/tfs/rm/ReleaseManagementCI.java
```diff
@@ -163,6 +241,36 @@ void CreateRelease(
             JSONObject object = new JSONObject(response);
             listener.getLogger().printf("Release Name: %s%n", object.getString("name"));
             listener.getLogger().printf("Release id: %s%n", object.getString("id"));
+
+            if (StringUtils.isNotBlank(this.destinationPath)) {
+                listener.getLogger().printf("Waiting for release to be finished...%n");
+                while (true) {
+                    String status = releaseManagementHttpClient.GetReleaseStatus(this.projectName, object.getString("id"));
+                    if (status.equals("finished")) {
```
kasubram: what if someone cancelled the release or deleted it. This will forever check the status. We should handle other status as well.

Also, what if there is a post approval which has not been approved for days. This will continue to check for that many days. We should also have a timeout to stop checking. 

For one of the above reason if the release is not succeeded and we stopped checking for status and no logs uploaded to jenkins workspace, in that case whenever the jenkins build page is opened we can check once again to see the current status of that release and update the jenkins workspace. is this achievable?

zackliu: I will check more status and add a time watcher. Do you know where to get documents about the all types of status. I can't find it.
For the last suggestion, I think it's hard to achieve. Because the environment may be changed a lot after build. The workspace may no longer exists after build (User may use dynamic slave and once the build is over, the slave will be deleted) or the configuration has changed. So deal with the old builds may cause unexpected issues. Thus, I think it's make sense to skip the unachievable logs.

jpricketMSFT:
@zackliu - The changes to the TfsConfiguration class seem fine, but the last commit is failing to build. The errors appear to be related to CheckStyle. Please correct the build errors so that we can merge this successfully.

zackliu:
@jpricketMSFT Could you please provide all the types of release status, I can't find it in docs. Thus I can fix the bug kasubram commented before.

jpricketMSFT:
@zackliu I am not familiar with the Release code. @kasubram - Could you please provide all the types of release status?

tfs/src/main/java/hudson/plugins/tfs/rm/ReleaseManagementCI.java
```diff
@@ -51,6 +71,58 @@ public ReleaseManagementCI(String collectionUrl, String projectName, String rele
...
+        this.releaseDefinitionName = releaseDefinitionName;
+        StandardUsernamePasswordCredentials credential = TeamCollectionConfiguration.findCredentialsById(credentialsId);
+        this.username = credential == null ? "" : credential.getUsername();
+        this.password = credential == null ? Secret.fromString("") : credential.getPassword();
+        this.credentialsId = credentialsId;
+        this.destinationPath = destinationPath;
+        this.isArchiveLog = isArchiveLog;
+    }
+
+    private Object readResolve() {
```
kasubram: shouldn't this be protected?

tfs/src/main/java/hudson/plugins/tfs/rm/ReleaseManagementCI.java
```diff
@@ -51,6 +71,58 @@ public ReleaseManagementCI(String collectionUrl, String projectName, String rele
...
+        if (StringUtils.isNotBlank(collectionUrl)
+                && StringUtils.isNotBlank(username)
```
kasubram: also if its VSTS and we are using PAT, the username will be empty.

tfs/src/main/java/hudson/plugins/tfs/rm/ReleaseManagementCI.java
kasubram: It is there in Microsoft.VisualStudio.Services.ReleaseManagement.WebApi.EnvironmentStatus
Possible status are,
Undefined, NotStarted, InProgress, Succeeded, Canceled, Rejected, Queued, Scheduled, PartiallySucceeded

tfs/src/main/java/hudson/plugins/tfs/rm/ReleaseManagementCI.java
```diff
@@ -163,6 +241,36 @@ void CreateRelease(
             JSONObject object = new JSONObject(response);
             listener.getLogger().printf("Release Name: %s%n", object.getString("name"));
             listener.getLogger().printf("Release id: %s%n", object.getString("id"));
+
+            if (StringUtils.isNotBlank(this.destinationPath)) {
+                listener.getLogger().printf("Waiting for release to be finished...%n");
+                while (true) {
+                    String status = releaseManagementHttpClient.GetReleaseStatus(this.projectName, object.getString("id"));
```
kasubram: change it to IsReleaseSucceded? and return a bool?

tfs/src/main/java/hudson/plugins/tfs/rm/ReleaseManagementHttpClient.java
```diff
@@ -119,6 +162,36 @@ private String ExecuteGetMethod(String url) throws ReleaseManagementException
...
+            if(status >= 300)
+            {
+                throw new ReleaseManagementException("Error occurred.%nStatus: " + status + "%nResponse: " + response + "%n");
+            }
+        }
+        catch(HttpException ex)
+        {
+            throw new ReleaseManagementException(ex);
+        }
+        catch(IOException ex)
```
kasubram: we are already dealing with the generic Exception. Do we need to catch the HttpException and IOException separately?

tfs/src/main/java/hudson/plugins/tfs/rm/ReleaseManagementHttpClient.java
```diff
@@ -57,6 +64,42 @@ public ReleaseArtifactVersionsResponse GetVersions(String project, List<Artifact
...
+            }
+        }
+        return "finished";
+    }
+
+    public List<Project> GetProjectItems() throws ReleaseManagementException
+    {
+        String url = this.accountUrl + "/_apis/projects?api-version=1.0";
+        String response = this.ExecuteGetMethod(url);
+        TeamProject teamProject = new Gson().fromJson(response, TeamProject.class);
```
kasubram: Can we do something like, 
values = new JSONObject(response).getString("value");
List<Project> projects = new Gson().fromJson(values, TeamProject[].class);

The teamproject object is not really representing the teamproject. The api returns a list of teamproject and a count. Looks like we want an intermediate object to represent that, but the final outcome we are looking is List<Project>. Looks like just parsing the value field is good enough.

tfs/src/main/java/hudson/plugins/tfs/rm/ReleaseManagementCI.java
```diff
@@ -163,6 +241,36 @@ void CreateRelease(
             JSONObject object = new JSONObject(response);
             listener.getLogger().printf("Release Name: %s%n", object.getString("name"));
             listener.getLogger().printf("Release id: %s%n", object.getString("id"));
+
+            if (StringUtils.isNotBlank(this.destinationPath)) {
```
kasubram: also can we wait for release completion only if isArchiveLog is set to true. with the current change it always waits for release completion whether we archive or not. If someone does not want to wait for release completion, he could uncheck isArchiveLog.

zackliu: In the current change, if user leave `destination path` blank, we won't wait for release completion. Do you means that if user uncheck `isArchiveLog`, whether he fills in `destination path` or not, we won't download logs for he.

tfs/src/main/java/hudson/plugins/tfs/rm/ReleaseManagementHttpClient.java
zackliu: I don't think it's useful to catch different types of exceptions but throw the same type in the end. However, I just follow the previous functions like `ExecuteGetMethod` or `ExecutePostmethod`.  So, do we need to also change these functions to just catch generic Exception.

zackliu:
@kasubram CI build failed because it said `File does not end with a newline` in `RemoteChangesetVersionCommand.java` and `TfsUserLookup.java`. But actually I didn't change these two files and I found there is one new line at the end of these files.
The Checkstyle passed in my local build, and I noticed a lot of build failed in CI because of this issue. So, any solutions?

tfs/src/main/java/hudson/plugins/tfs/TeamCollectionConfiguration.java
```diff
@@ -246,6 +260,32 @@ static StandardUsernamePasswordCredentials findCredentialsById(final String cred
         return result;
     }
 
+    public static String setCredentials(final String hostName, String username, String password) {
+        final HostnameRequirement requirement = new HostnameRequirement(hostName);
+
+        if (!SystemCredentialsProvider.getInstance().getDomainCredentialsMap().containsKey(requirement)) {
+            generateDomain(hostName);
+        }
```
kasubram: leave a space after the brace

tfs/src/main/java/hudson/plugins/tfs/rm/ReleaseManagementCI.java
```diff
@@ -227,5 +280,79 @@ public String getDisplayName()
...
+                            collectionUrl.toLowerCase().replaceFirst(".visualstudio.com", ".vsrm.visualstudio.com"),
+                            username,
+                            password);
+
+            try {
+                List<ReleaseDefinition> releaseDefinitions = releaseManagementHttpClient.GetReleaseDefinitions(projectName);
+                for (ReleaseDefinition releaseDefinition : releaseDefinitions) {
+                    listBoxModel.add(releaseDefinition.getName());
+                }
+            } catch (ReleaseManagementException ignore) {
```
kasubram: please log

tfs/src/main/java/hudson/plugins/tfs/rm/ReleaseManagementHttpClient.java
```diff
@@ -57,6 +65,29 @@ public ReleaseArtifactVersionsResponse GetVersions(String project, List<Artifact
         String response = this.ExecutePostmethod(url, body);
         return new Gson().fromJson(response, ReleaseArtifactVersionsResponse.class);
     }
+
+    public void GetReleaseLogs(String project, String releaseId, BuildListener listener) throws ReleaseManagementException
+    {
+        String url = this.accountUrl + project + "/_apis/release/releases/" + "2" + "/logs?api-version=4.0-preview.2";
```
kasubram: aren't you supposed to use the releaseId instead of hardcoding "2"?

tfs/src/main/java/hudson/plugins/tfs/rm/ReleaseManagementHttpClient.java
```diff
@@ -57,6 +65,29 @@ public ReleaseArtifactVersionsResponse GetVersions(String project, List<Artifact
...
+        InputStream inputStream = this.ExecuteGetMethodAsStream(url);
+
+        try {
+            FileOutputStream out = new FileOutputStream("/home/chenyl/tmp.zip");
```
kasubram: please remove this home path. You need use File.CreateTempFile.

tfs/src/main/java/hudson/plugins/tfs/rm/ReleaseManagementHttpClient.java
```diff
@@ -57,6 +65,29 @@ public ReleaseArtifactVersionsResponse GetVersions(String project, List<Artifact
...
+            IOUtils.copy(inputStream, out);
+            out.close();
+        } catch (Exception e) {
+
```
kasubram: Please log exception

tfs/src/main/java/hudson/plugins/tfs/rm/ReleaseManagementHttpClient.java
zackliu: Sorry, I forgot to clean some debugging codes.

tfs/src/main/java/hudson/plugins/tfs/rm/ReleaseManagementHttpClient.java
kasubram: I am not sure how we ended up having that code to catch different exception but finally throw the same exception. Please fix  your method just to handle generic exception. Get and Post method I will fix it later.

zackliu:
@kasubram As we discussed, I removed archiving logs and add a link to VSTS logs page.
I updated the summary and screenshots, thus you can have a brief understanding.

tfs/src/main/java/hudson/plugins/tfs/rm/ReleaseManagementCI.java
```diff
@@ -1,38 +1,55 @@
...
 
 /**
  * @author Ankit Goyal
  */
-public class ReleaseManagementCI extends Notifier{
+public class ReleaseManagementCI extends Notifier implements Serializable {
+
+    private static final long serialVersionUID = -760016860995557L;
+
+    private static final long RELEASE_TIMEOUT_MINUTES = 120;
```
kasubram: please remove the unused variables

tfs/src/main/java/hudson/plugins/tfs/rm/ReleaseManagementCI.java
```diff
@@ -163,6 +233,9 @@ void CreateRelease(
             JSONObject object = new JSONObject(response);
             listener.getLogger().printf("Release Name: %s%n", object.getString("name"));
             listener.getLogger().printf("Release id: %s%n", object.getString("id"));
+            String logLink = this.collectionUrl + this.projectName
```
kasubram: can we call it releaseUrl?

tfs/src/main/java/hudson/plugins/tfs/rm/ReleaseManagementCI.java
```diff
@@ -163,6 +233,9 @@ void CreateRelease(
...
+                    + "/_release?releaseId=" + object.getString("id") + "&_a=release-logs";
```
kasubram: Please do not construct the url by hand. In the release object under _link property you will notice "self" and 'web" keys. Use the url under the "web" key.

tfs/src/main/resources/hudson/plugins/tfs/rm/LogSummaryAction/summary.jelly
```diff
@@ -0,0 +1,6 @@
+<?jelly escape-by-default='true'?>
```
kasubram: Can we call it release summary?

tfs/src/main/java/hudson/plugins/tfs/rm/LogSummaryAction.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Show release logs link in build page.
+ */
+
+@ExportedBean
+public class LogSummaryAction extends InvisibleAction {
+    private String projectName;
+    private int buildNo;
+    private String logLink;
```
kasubram: releaseLink?

tfs/src/main/java/hudson/plugins/tfs/rm/LogSummaryAction.java
```diff
@@ -0,0 +1,37 @@
...
+
+import hudson.model.InvisibleAction;
+import org.kohsuke.stapler.export.ExportedBean;
+
+/**
+ * Show release logs link in build page.
+ */
+
+@ExportedBean
+public class LogSummaryAction extends InvisibleAction {
```
kasubram: can we call it ReleaseSummaryAction?

tfs/src/main/java/hudson/plugins/tfs/rm/ReleaseManagementCI.java
```diff
@@ -163,6 +232,8 @@ void CreateRelease(
             JSONObject object = new JSONObject(response);
```
kasubram: are we not converting to ReleaseDetails? if so then ReleaseDetails is not used anymore. If its not used anymore please remove the ReleaseDetails type.

tfs/src/main/java/hudson/plugins/tfs/rm/ReleaseSummaryAction.java
```diff
@@ -0,0 +1,37 @@
+package hudson.plugins.tfs.rm;
+
+import hudson.model.InvisibleAction;
+import org.kohsuke.stapler.export.ExportedBean;
+
+/**
+ * Show release logs link in build page.
```
kasubram: show release link..

tfs/src/main/java/hudson/plugins/tfs/rm/ReleaseManagementCI.java
```diff
@@ -227,5 +298,91 @@ public String getDisplayName()
...
+                        new ReleaseManagementHttpClient(
+                                collectionUrl.toLowerCase(),
+                                username,
+                                password);
+
+                List<Project> projects = releaseManagementHttpClient.GetProjectItems();
+                for (Project project : projects) {
+                    listBoxModel.add(project.getName());
+                }
+            } catch (ReleaseManagementException ignored) {
```
kasubram: please log.. otherwise we will have tough time to debug..

tfs/src/main/java/hudson/plugins/tfs/rm/ReleaseManagementCI.java
```diff
@@ -51,6 +67,53 @@ public ReleaseManagementCI(String collectionUrl, String projectName, String rele
...
+                    if ((StringUtils.isBlank(username) || credential.getUsername().equals(username))
+                            && credential.getPassword().getPlainText().equals(password.getPlainText())) {
+                        this.credentialsId = credential.getId();
+                        return this;
+                    }
+                }
+                this.credentialsId
+                        = TeamCollectionConfiguration.setCredentials(hostName, username, password.getPlainText());
+            } catch (Exception ignore) {
+
```
kasubram: please log

zackliu:
@kasubram Seems the PR has passed the CI. Could we merge it?

tfs/src/main/java/hudson/plugins/tfs/TeamCollectionConfiguration.java
```diff
@@ -246,6 +272,32 @@ static StandardUsernamePasswordCredentials findCredentialsById(final String cred
...
+                "Generated for " + username,
+                username,
+                password
+        );
+        SystemCredentialsProvider.getInstance().getDomainCredentialsMap().get(domain).add(newCredential);
+
+        try {
+            SystemCredentialsProvider.getInstance().save();
+        } catch (IOException ignore) {
+
```
GitHubSriramB: add a warning log?

PR #190, opened 2017-12-11
Users/kej/update changeset url

kelliejos:
I updated the URL we craft to point back to a changeset in TFS/VSTS to use the valid format of `projectname/_versionControl/changeset/10210 ` instead of `projectname/_versionControl/changeset?id=10210`. This fixes [JENKINS-48341](https://issues.jenkins-ci.org/browse/JENKINS-48341)

I also removed the `//CHECKSTYLE:OFF` annotation and made the necessary code changes to get it to pass.

### DIFF
--- a/tfs/src/main/java/hudson/plugins/tfs/browsers/TeamSystemWebAccessBrowser.java
+++ b/tfs/src/main/java/hudson/plugins/tfs/browsers/TeamSystemWebAccessBrowser.java
@@ -1,4 +1,3 @@
-//CHECKSTYLE:OFF
 package hudson.plugins.tfs.browsers;
 
 import hudson.Extension;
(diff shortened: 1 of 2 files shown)


tfs/src/main/java/hudson/plugins/tfs/browsers/TeamSystemWebAccessBrowser.java
```diff
@@ -62,13 +62,10 @@ private String getBaseUrlString(ChangeSet changeSet) throws MalformedURLExceptio
     public URL getChangeSetLink(final ChangeSet changeSet) throws IOException {
         final String baseUrlString = getBaseUrlString(changeSet);
         final URL baseUrl = new URL(baseUrlString);
```
jpricketMSFT: When did the URL change? Do we need to be backwards compatible?

kelliejos: The URL hasn't changed, I believe this URL has always been incorrect. We should be safe.

PR #192, opened 2018-01-12
Update POMs and cleanup the dependencies/code

oleg-nenashev:
I was working on PCT for JEP-200, and finally I discovered that the plugin build does not work there. In this PR i tried to cleanup the POMs and code a bit.

- [x] Core dependency is bumped to the version required by current Pipeline dependencies
- [x] Use Parent POM 3.2
- [x] Get rid of explicit Core/War dependencies in TFS Plugin. They were suppressing some analysis, and generally it's not required since parent POM inherits from Plugin POM
- [x] Fix compilation failures in the code. For me it was failing even in the stock version, but please double-check

@reviewbybees

### DIFF
--- a/tfs/pom.xml
+++ b/tfs/pom.xml
@@ -11,6 +11,12 @@
   <name>Team Foundation Server Plug-in</name>
   <version>5.126.0-SNAPSHOT</version>
   <url>http://wiki.jenkins-ci.org/display/JENKINS/Team+Foundation+Server+Plugin</url>
+
+  <properties>
+    <!-- TODO: enable once FindBugs issues are fixed -->
+    <findbugs.failOnError>false</findbugs.failOnError>
+  </properties>
+
   <licenses>
     <license><name>MIT license</name><comments>All source code is under the MIT license.</comments></license>
     <license>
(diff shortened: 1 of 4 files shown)


tfs/src/main/java/hudson/plugins/tfs/telemetry/TelemetryContextInitializer.java
```diff
@@ -111,7 +111,8 @@ private String getUserId() {
         final String userName = getSystemProperty(SYS_PROP_USER_NAME);
         final String fakeUserId = MessageFormat.format("{0}@{1}", userName, computerName);
 
-        return DigestUtils.sha1Hex(fakeUserId);
+        //FIXME: was sha1Hex, but this is available only in commons codec 1.7. TFS SDK 14.0.3 bundles older version
```
jpricketMSFT: I don't think this needs to be fixed (i.e. you can remove the comment). The idea here is to keep the user id unique to the telemetry server, but not allow anyone to actually identify the user. So, as long as the user id is obfuscated, there shouldn't be a problem.

tfs/pom.xml
```diff
@@ -11,6 +11,12 @@
   <name>Team Foundation Server Plug-in</name>
   <version>5.126.0-SNAPSHOT</version>
   <url>http://wiki.jenkins-ci.org/display/JENKINS/Team+Foundation+Server+Plugin</url>
+
+  <properties>
+    <!-- TODO: enable once FindBugs issues are fixed -->
+    <findbugs.failOnError>false</findbugs.failOnError>
```
kelliejos: Is this necessary? I hesitate to turn it off.

oleg-nenashev: With the current configuration you will be getting FindBugs warnings in the build log. They just won't fail the build.

PR #193, opened 2018-03-07
doc: remove wrong environment replacement

jhinrichsen:
The documentation in here states that ${ENV} will be replaced with the corresponding environment variable. This is not true, no such replacement takes place:

    Creating workspace '${TFS_ID}-tfstest' owned by ...

 This is consistent with the documentation of the TFS Plugin itself (https://github.com/jenkinsci/tfs-plugin/blob/master/README.md) which mentions  only ${JOB_NAME}, ${USER_NAME}, and ${NODE_NAME}.


jhinrichsen:
@jpricketMSFT Can you please review this PR?

jhinrichsen:
How can removing one line of HTML break the CI? The error is 

/home/jenkins/workspace/Plugins_tfs-plugin_PR-193-2SSF2HRRDQBRX55DZWQUIXD7N275LKZOQBC7S72PXSJIBEE5CZSA/tfs/src/main/java/hudson/plugins/tfs/util/KeyValueTextReader.java:17: error: Line has trailing spaces.

I have not even touched that file. Seems as if the new Checkstyle rules broke the CI build.

PR #194, opened 2018-03-22
Revert "Allow users to map sub-folders of cloaked paths"

kelliejos:
Reverting jenkinsci/tfs-plugin#181 until we can verify that polling is still functioning properly and we can fix the tests.


kelliejos:
Abandoning this PR as it inadvertently reverts some changes we want to keep.

PR #195, opened 2018-03-22
Revert "Allow users to map sub-folders of cloaked paths (#181)"

kelliejos:
This reverts commit f6fe6de50a52e4c0fe3419b996bdc8204c912210. Reverting
until we can ensure that polling has not been affected and we can fix
teh functional tests.


tfs/src/main/java/hudson/plugins/tfs/model/WorkspaceConfiguration.java
```diff
@@ -132,35 +121,6 @@ else if (cloakedPaths.size() != other.cloakedPaths.size())
             return false;
         else if (!cloakedPaths.containsAll(other.cloakedPaths))
             return false;
-        if (mappedPaths == null) {
```
jpricketMSFT: FYI, my guess is the problem with polling might be in here. This is the most complicated "equals" method I have ever seen. We probably need to revisit this entire method.

PR #198, opened 2018-04-27
Recreating the instanceReference object correctly for every artifact,…

kasubram:
… otherwise same version is applied for all the available artifacts.


kasubram:
@kelliejos Can you please merge this aswell?

PR #199, opened 2018-05-29
Restore Support for TFS 2017/2018 in Build Status

MartinPayne:
Use REST API version "3.0-preview.1" when posting commit or PR status
back to TFS, because version "4.1-preview" is not supported by TFS 2017
or 2018.

Resolves JENKINS-51579.

### DIFF
--- a/tfs/src/main/java/hudson/plugins/tfs/util/TeamRestClient.java
+++ b/tfs/src/main/java/hudson/plugins/tfs/util/TeamRestClient.java
@@ -280,7 +280,7 @@ public class TeamRestClient {
 
     public TeamGitStatus addPullRequestStatus(final PullRequestMergeCommitCreatedEventArgs args, final TeamGitStatus status) throws IOException {
 
-        final QueryString qs = new QueryString(API_VERSION, "4.1-preview");
+        final QueryString qs = new QueryString(API_VERSION, "3.0-preview.1");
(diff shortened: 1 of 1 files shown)


tfs/src/main/java/hudson/plugins/tfs/util/TeamRestClient.java
```diff
@@ -280,7 +280,7 @@ public void addHyperlinkToWorkItem(final int workItemId, final String hyperlink)
 
     public TeamGitStatus addPullRequestStatus(final PullRequestMergeCommitCreatedEventArgs args, final TeamGitStatus status) throws IOException {
 
-        final QueryString qs = new QueryString(API_VERSION, "4.1-preview");
+        final QueryString qs = new QueryString(API_VERSION, "3.0-preview.1");
```
jpricketMSFT: Can you use a released version? The APIs are on 5.0 now. It seems like 3.0 should not need the preview tag.

MartinPayne: According to the documentation, the commit status API request is no longer in preview but the pull request status still is. However, for compatibility with TFS 2017 and 2018, "3.0-preview.1" looks to be the only supported version. Any other version gives a "404 - Not Found" HTTP response.

PR #200, opened 2018-06-18
Improve VSTS Service Hook integration

mrflow:
Deduplicate code paths used for feeding back status to VSTS/TFS.

Further, set the "name" and "genre" fields to sane values:

VSTS uses these fiels for configuration of branch policies. Empty
strings are not permitted, thus this feature was broken on the part of
tfs-plugin.

Use the Jenkins root-URL as 'genre' and the job-name as 'name'.

### DIFF
--- a/tfs/src/main/java/hudson/plugins/tfs/model/TeamGitStatus.java
+++ b/tfs/src/main/java/hudson/plugins/tfs/model/TeamGitStatus.java
@@ -4,7 +4,9 @@ package hudson.plugins.tfs.model;
 import hudson.model.Job;
 import hudson.model.Result;
 import hudson.model.Run;
+import jenkins.model.Jenkins;
 import net.sf.json.JSONObject;
+import org.apache.commons.lang.StringUtils;
 
 import javax.annotation.Nonnull;
 import java.util.Collections;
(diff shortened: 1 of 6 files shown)


mrflow:
I also implemented abortion of currently running builds with the same pull-request as a cause.

tfs/src/main/java/hudson/plugins/tfs/model/TeamGitStatus.java
```diff
@@ -36,19 +38,32 @@ public static TeamGitStatus fromRun(@Nonnull final Run<?, ?> run) {
...
+        status.state = GitStatusState.Pending;
+        status.description = "Jenkins Job " + job.getDisplayName() + " queued";
+        status.targetUrl = job.getAbsoluteUrl();
+        status.context = getStatusContext(job);
+        return status;
+    }
+
+    private static GitStatusContext getStatusContext(@Nonnull final Job job) {
+        final String instanceUrl = StringUtils.stripEnd(Jenkins.getInstance().getRootUrl(), "/");
+        final String projectDisplayName = job.getDisplayName();
```
mrflow: This should be getFullDisplayName instead.

mrflow: or probably not, because display names might not be stable. What we need is something that takes into account the directory structure that is possible with jenkins plugins.

mrflow:
@kelliejos IMHO, the pull-request is ready for merging.

kelliejos:
Merged. Thanks again for your contribution!

PR #208, opened 2018-08-20
Adding release webhook configuration to global configuration

kasubram:


tfs/src/main/java/hudson/plugins/tfs/TeamEventsEndpoint.java
```diff
@@ -225,6 +227,16 @@ public void doConnect(
         dispatch(request, response, body);
     }
 
+    @RequirePOST
+    public void doRmwebhook(
+            final StaplerRequest request,
+            final StaplerResponse response,
+            @StringBodyParameter @Nonnull final String body) {
+        // Send telemetry
+        TelemetryHelper.sendEvent("team-events-rmwebhook", new TelemetryHelper.PropertyMapBuilder().build());
```
GitHubSriramB: where is this telemetry captured? what details get logged in the telemetry?

tfs/src/main/java/hudson/plugins/tfs/TeamEventsEndpoint.java
```diff
@@ -62,6 +63,7 @@
         eventMap.put("gitPullRequestMerged", new GitPullRequestMergedEvent.Factory());
         eventMap.put("gitPush", new GitPushEvent.Factory());
         eventMap.put("connect", new ConnectHookEvent.Factory());
+        eventMap.put("rmwebhook", new ConnectReleaseWebHookEvent.Factory());
```
GitHubSriramB: rmWebhook?

tfs/src/main/java/hudson/plugins/tfs/TeamPluginGlobalConfig.java
```diff
@@ -108,6 +118,9 @@ public void setUserAccountMapper(UserAccountMapper userAccountMapper) {
     public boolean configure(StaplerRequest req, JSONObject json) throws FormException {
```
GitHubSriramB: is this called when the project is loaded in UI?

tfs/src/main/java/hudson/plugins/tfs/TeamEventsEndpoint.java
```diff
@@ -225,6 +227,16 @@ public void doConnect(
         dispatch(request, response, body);
     }
 
+    @RequirePOST
+    public void doRmwebhook(
```
GitHubSriramB: is this called for every REST API call with rmwebhook in the resource? is the name indicative of the registration?

tfs/src/main/java/hudson/plugins/tfs/TeamPluginGlobalConfig.java
kasubram: When the global setting page is loaded

tfs/src/main/java/hudson/plugins/tfs/TeamEventsEndpoint.java
kasubram: This is root for all rmwebhook related changes.. we have register/link not sure what generic name to use. Just using RmWebHook looks good to me as its the root.

tfs/src/main/java/hudson/plugins/tfs/TeamEventsEndpoint.java
kasubram: Fixed.

PR #210, opened 2018-09-19
tell if branch is missing

CSchulz:
Improve the error message if the job can be found but the source branch is missing


CSchulz:
I am not sure why checkstyle is failing:

> \tfs\src\main\java\hudson\plugins\tfs\JenkinsEventNotifier.java:0: error: File does not end with a newline.
\tfs\src\main\java\hudson\plugins\tfs\rm\ReleaseWebHookHelper.java:0: error: File does not end with a newline.

PR #211, opened 2018-10-01
BuildCommand: abort a build only if it was caused by the same PR

mrflow:
There was a bug that caused a running job, triggered by a merge-commit Webhook,
being canceled regardless of what the running job's build cause was.

This fixes the misbehavior by just canceling builds that have the same
Pullrequest as a build cause.


kelliejos:
Good catch! Thanks for submitting the fix.

PR #215, opened 2018-11-12
Implemented switch for 'local' or 'server' workspaces.

ytterx:
Under "ADVANCED" I added an extra checkbox if the workspace should be 'Local' or 'Server'.
The difference is described at https://docs.microsoft.com/en-us/azure/devops/repos/tfvc/decide-between-using-local-server-workspace
But one very important detail is that the file attributes also change with this option upon checkout. So for us, all workspaces must therefore be a 'Local' workspace.


ytterx:
Can someone take a look at the changes?

ytterx:
@keljos @jpricket There are over 10 Pull request created in the last year but nobody has looked at them, let alone accept them (even older ones which are looked at and "accepted" are not merged). Any plan for this?

PR #216, opened 2018-11-27
Persist Pull-Request details

fbaeuerle:
Make sure to save the pull-request trigger details persistently for being able
to still display them after restarting Jenkins.

Fixes: JENKINS-54883


mrflow:
It's a bit messy because there's some redundancy between TeamPullRequestMergedDetailsAction and PullRequestParameterAction. The former is used for displaying trigger information in the Jenkins UI. To me it pretty much looks like there is no real necessity for PullRequestParameterAction.

I think the plugin needs some heavy refactoring by someone who actually understands all the ways a build can be triggered using the plugin.

PR #218, opened 2018-11-29
Implement PR iteration status for generic build trigger

fbaeuerle:
The REST API does not make this particularly easy, so rely on comparing
the team-events source commit with the iteration statuses source commit.

Fixes: JENKINS-51973


mrflow:
It must be sufficient to compare the source commit, otherwise no iteration will be found when restarting a merge in azure devops.

PR #220, opened 2019-02-04
Removed the javax.xml.bind.DatatypeConverter as part of enabling Java 9

AniketSK:
Since the javax.xml.bind package is removed in Java 9, it would help if
this package was removed from the code.
This was replaced with Apache Commons and not java.util.Base64 because
the encode function there requires compilation with java 8.


batmat:
cc @jenkinsci/java11-support

batmat:
Relates to https://issues.jenkins-ci.org/browse/JENKINS-55941 FYI @kelliejos @jpricketMSFT @kasubram

PR #221, opened 2019-02-09
Account for bad repo path while searching projects

shimpossible:
Any projects with Git repo paths that can not be converting to a URI stop all other repos from being scanned

The exception looks like the following

java.lang.IllegalArgumentException: Illegal character in opaque part at index 2: c:\repo\test_repo
        at java.net.URI.create(URI.java:852)
        at hudson.plugins.tfs.util.UriHelper.areSameGitRepo(UriHelper.java:133)
        at hudson.plugins.tfs.model.AbstractHookEvent.pollOrQueueFromEvent(AbstractHookEvent.java:302)
        at hudson.plugins.tfs.model.GitPushEvent.perform(GitPushEvent.java:55)
        at hudson.plugins.tfs.TeamEventsEndpoint.innerDispatch(TeamEventsEndpoint.java:171)
        at hudson.plugins.tfs.TeamEventsEndpoint.dispatch(TeamEventsEndpoint.java:141)
        at hudson.plugins.tfs.TeamEventsEndpoint.doGitPush(TeamEventsEndpoint.java:216)

### DIFF
--- a/tfs/src/main/java/hudson/plugins/tfs/model/AbstractHookEvent.java
+++ b/tfs/src/main/java/hudson/plugins/tfs/model/AbstractHookEvent.java
@@ -299,9 +299,14 @@ public abstract class AbstractHookEvent {
                     for (final RemoteConfig repository : git.getRepositories()) {
                         boolean repositoryMatches = false;
                         for (final URIish remoteURL : repository.getURIs()) {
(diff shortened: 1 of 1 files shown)


shimpossible:
Rather than try to parse the path in UriHelper, I've opted to just catch the exception in AbstractHookEvent and move to the next repo URL.

The error is logged so people can see what is going on

PR #222, opened 2019-02-09
 prevent nullpointer in SafeParametersAction

shimpossible:
In Jenkins 1.642 the following NullPointerException happens when pushing an event

SEVERE: Error while performing reaction to 'gitPush' event.
java.lang.NullPointerException
	at hudson.model.ParametersAction.buildEnvVars(ParametersAction.java:87)
	at hudson.plugins.git.util.GitUtils.addEnvironmentContributingActionsValues(GitUtils.java:292)
	at hudson.plugins.git.util.GitUtils.getPollEnvironment(GitUtils.java:278)
	at hudson.plugins.git.GitSCM.compareRemoteRevisionWithImpl(GitSCM.java:599)
	at hudson.plugins.git.GitSCM.compareRemoteRevisionWith(GitSCM.java:574)
	at hudson.scm.SCM.compareRemoteRevisionWith(SCM.java:381)
	at hudson.scm.SCM.poll(SCM.java:398)
	at hudson.model.AbstractProject._poll(AbstractProject.java:1453)
	at hudson.model.AbstractProject.poll(AbstractProject.java:1356)
	at hudson.plugins.tfs.TeamPushTrigger$Runner.runPolling(TeamPushTrigger.java:109)
	at hudson.plugins.tfs.TeamPushTrigger$Runner.run(TeamPushTrigger.java:142)

Later versions of Jenkins have deprecated buildEnvVars and also check for a NullPointer to prevent this.

Tracking it down shows the problem is ParemtersAction this.paremters is null.  SafeParametersAction was not correctly calling its parent constructor to pass in the parameters list.

### DIFF
--- a/tfs/src/main/java/hudson/plugins/tfs/SafeParametersAction.java
+++ b/tfs/src/main/java/hudson/plugins/tfs/SafeParametersAction.java
@@ -24,6 +24,7 @@ public class SafeParametersAction extends ParametersAction {
     private List<ParameterValue> parameters;
 
     public SafeParametersAction(final List<ParameterValue> parameters) {
+        super(parameters);
         this.parameters = parameters;
     }
 


shimpossible:
You could also override the deprecated function and check for null, but calling the constructor looks to be the simplest

```
    @Override
    public void buildEnvVars(final hudson.model.AbstractBuild<?, ?> build, final hudson.EnvVars env) {
       if(this.paremters == null) return;
        for (ParameterValue p : parameters) {
            if (p == null) {
                continue;
            }
            p.buildEnvironment(build, env);
        }
    }

```

shimpossible:
After looking into this more, this may not be the correct fix.   I'm not able to get the crash to happen consistently.  Its something to do with deserializing, and possibly the parent class also having 'parameters' field.   Unfortunately I can't duplicate the crash anymore and the XML has both a 
<parameters defined-in="hudson.plugins.tfs.SafeParametersAction"> and  <parameters class="java.util.Arrays$ArrayList">

It was fixed in later versions by ensuring ParametersAction didn't have a null pointer.
https://issues.jenkins-ci.org/browse/JENKINS-39495

shimpossible:
Using readResolve to verify ParametersAction is initialized correctly seems like the only sure way.
the NullPointer came back this morning.  The build.xml is the same as yesterday, but did not appear to be unmarshelling correctly.

 ```
   <hudson.plugins.tfs.SafeParametersAction plugin="tfs@5.142.0-SNAPSHOT">
      <parameters class="java.util.Arrays$ArrayList">
        <a class="hudson.model.ParameterValue-array"/>
      </parameters>
      <parameters defined-in="hudson.plugins.tfs.SafeParametersAction">
        <hudson.model.StringParameterValue>
          <name>vstsRefspec</name>
          <value>+refs/heads/*:refs/remotes/origin/*</value>
        </hudson.model.StringParameterValue>
        <hudson.model.StringParameterValue>
          <name>vstsBranchOrCommit</name>
          <value>8d03babcd6fafbc3cc9814e345bdc47155fe350f</value>
        </hudson.model.StringParameterValue>
      </parameters>
```

PR #224, opened 2019-03-07
Add support for dev.azure.com style URLs

omusavi:
Using the old visualstudio.com style URL still work to find and configure the project collection, but when trying to set up a Post-Build service callback back to Azure DevOps, the plugin currently fails with the following error:
`ERROR: There is no team project collection configured for the URL 'https://dev.azure.com/<collection>/'.
Please go to Jenkins > Manage Jenkins > Configure System and then add a Team Project Collection with a Collection URL of 'https://dev.azure.com/<collection>/'.`

Seems like at some level, the server must respond with it's own URL which is different than what the plugin expects. On top of that, new users of Azure DevOps may not know about the old URL format and putting in `dev.azure.com/<collection name>` gives a validation error.

This PR modified the validation methods that check the hostname for *.visualstudio.com to also accept dev.azure.com as a hostname.

### DIFF
--- a/tfs/src/main/java/hudson/plugins/tfs/TeamCollectionConfiguration.java
+++ b/tfs/src/main/java/hudson/plugins/tfs/TeamCollectionConfiguration.java
@@ -172,7 +172,7 @@ public class TeamCollectionConfiguration extends AbstractDescribableImpl<TeamCol
     }
 
     static FormValidation checkTeamServices(final URI uri) {
-        if (UriHelper.hasPath(uri)) {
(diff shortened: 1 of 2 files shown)


omusavi:
Note that I am not an expert in this plugin, but this change was enough to unblock our scenario on our local Jenkins instance. If there are additional places where changes should be made let me know! Seems in general there should probably be some renaming work done to go from TFS -> AzDO but thats a pretty big item :)

jdstapleton:
I created a jira issue linking this ticket: [JENKINS-61718](https://issues.jenkins-ci.org/browse/JENKINS-61718)

PR #231, opened 2020-02-15
Using full name of the project to make sure rm webhook trigger works …

kasubram:
…for projects under the folder


UmmerS:
@kasubram
please release the latest TFS plugin as we are unable to use old plugin in latest jenkins 
Thanks in advance

PR #239, opened 2020-10-21
update outdated path of "TFS SDK for Java"

trivalik:


trivalik:
Can you please merge this. I changed only plain text.

ianwilliams1 (changes requested):
Reference URL should be: https://github.com/microsoft/team-explorer-everywhere
Download URL should be: https://github.com/microsoft/team-explorer-everywhere/releases

trivalik:
@ianwilliams1  not sure what you mean, should I change the link text to your suggested "Reference URL"?

I found that the latest URL what is used in this plugin can be found here:

http://www.microsoft.com/en-us/download/details.aspx?id=47727

But in this version are API used which in current version not anymore present.

